PR #28, opened 2016-05-26
Give access to ID token fields

jorunfa:


README.md
```diff
@@ -216,10 +216,13 @@ method. When the token has expired a new set of tokens can be requested using
 Access tokens can be used to access resources on your resource server. Please refer to the document
 about [scope tokens](http://docs.telenordigital.com/connect/id/scope.html) for more details.
 
-### Retrieving the logged in user's ID
+### Retrieving the logged in user's ID, email or phone number
```
ulfcarlin: "Retrieving information about the logged in user"?

README.md
```diff
@@ -216,10 +216,13 @@ method. When the token has expired a new set of tokens can be requested using
...
+
+If you supply email and phone number as scope and claims you can access these fields on the user
```
ulfcarlin: Do you mean "requesting email and phone scope values or claims"?

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -326,7 +327,20 @@ private static void loadConnectConfig(Context context) {
      */
     public static String getSubjectId() {
```
ulfcarlin: Have you considered deprecating this method?

connect/src/com/telenor/connect/id/IdToken.java
```diff
@@ -4,13 +4,24 @@
...
 import com.telenor.connect.ConnectException;
 
-import java.io.Serializable;
 import java.text.ParseException;
+import java.util.Map;
 
-public class IdToken implements Serializable {
+public class IdToken {
+
+    private static final String AUTHENTICATED_USERNAME = "td_au";
```
ulfcarlin: "AUTHENTICATION_USERNAME" to be consistent

connect/src/com/telenor/connect/id/IdToken.java
```diff
@@ -34,11 +54,36 @@ public String getSubject() {
...
+    }
+
+    public String getPhoneNumber() {
+        return phoneNumber;
+    }
+
+    public Boolean getPhoneNumberVerified() {
+        return phoneNumberVerified;
+    }
+
```
ulfcarlin: Consider exposing the "name" claim as well (and also document it in README.md)

connect/src/com/telenor/connect/id/IdToken.java
```diff
@@ -23,7 +34,16 @@ public IdToken(String serializedSignedJwt) {
                     "Could not parse saved id token. idToken=" + this, e);
         }
 
+        final Map<String, Object> cc = jwtClaimsSet.getCustomClaims();
```
ulfcarlin: Move the declaration closer to the first usage, that is, one line down. Consider avoiding short and cryptic variable names, even in such a limited scope: cc => customClaims.

connect/src/com/telenor/connect/id/IdToken.java
```diff
@@ -23,7 +34,16 @@ public IdToken(String serializedSignedJwt) {
...
         subject = jwtClaimsSet.getSubject();
+        authenticationUsername = cc.containsKey(AUTHENTICATED_USERNAME)
+                ? (String) cc.get(AUTHENTICATED_USERNAME) : null;
+        email = cc.containsKey(EMAIL) ? (String) cc.get(EMAIL) : null;
+        emailVerified
+                = cc.containsKey(EMAIL_VERIFIED) ? (Boolean) cc.get(EMAIL_VERIFIED) : null;
```
ulfcarlin: Consider using same line break as for other claims

README.md
jorunfa: Fixed.

README.md
jorunfa: Better now?

connect/src/com/telenor/connect/ConnectSdk.java
jorunfa: Was thinking about it. Thanks for the push.

connect/src/com/telenor/connect/id/IdToken.java
jorunfa: Fixed.

connect/src/com/telenor/connect/id/IdToken.java
jorunfa: Fixed.

connect/src/com/telenor/connect/id/IdToken.java
jorunfa: Moved it down. I think I'll use to option to squash the commits for this branch when I merge.

connect/src/com/telenor/connect/id/IdToken.java
jorunfa: Better now?

README.md
```diff
@@ -216,10 +216,15 @@ method. When the token has expired a new set of tokens can be requested using
 Access tokens can be used to access resources on your resource server. Please refer to the document
 about [scope tokens](http://docs.telenordigital.com/connect/id/scope.html) for more details.
 
-### Retrieving the logged in user's ID
+### Retrieving information about the logged in user
+
+If you request email, phone, profile and/or name scope tokens and claims, you can access these
```
ulfcarlin: "name" is only a claim, not a scope token. "profile" is only a scope token and not a claim. "email" and "phone" are both scope tokens and claims.

Maybe you can rewrite to something like the following?

"If you request user claims like email, phone, and name using either scope tokens or the claims parameter, (…). If both email and phone claims have been requested, we will also provided the username used for the authentication in the ID token.

jorunfa: Better now, or still not quite as you wanted it?

connect/src/com/telenor/connect/id/Claims.java
```diff
@@ -8,6 +8,7 @@
 
     public static final String OPENID = "openid";
     public static final String PROFILE = "profile";
```
ulfcarlin: I don't understand this. "openid" is not a claim and we do not support the "profile" claim yet (only the profile scope which "maps" to several different claims, including the profile and name claims). I think I need to take a closer look at this code.

jorunfa: I might have misinterpreted the docs?:

> The available claims to request have the same values as the available scope values

http://docs.telenordigital.com/apis/connect/id/authentication.html

ulfcarlin: The documentation is unfortunately not accurate, since claims are separate from and more fine-grained than scope values. I will look into improving the docs. I'm sorry I forgot to do the review for UXUI-558 while being on parental leave.

We do only support the claims listed in https://gerrit.c.skunk-works.no/gitweb/?p=comoyoauth.git;a=blob;f=oauth-protocol/src/main/java/com/comoyo/oauth/protocol/Claim.java;hb=refs/heads/master, see also the documentation for the userinfo endpoint: http://docs.telenordigital.com/apis/connect/id/authentication.html#authorization-server-user-information.

Also, in the authorization server, we take some liberties with respect to how we interpret the claims request. More specifically, we ignore whether they are requested for userinfo or id_token and include them both places. We also map fine-grained claims to more course-grained scope values internally, so we are sometimes returning more claims than were requested.

It would be nice if you could update the claims functionality in the SDK to only include the supported claims, and possibly if you could support using both userinfo and id_token in the claims request? Consider doing this in a separate task/commit.

ulfcarlin:
LGTM

connect/src/com/telenor/connect/id/Claims.java
ulfcarlin: In addition, it would have been nice to make the SDK doc about claims clearer. The claims set end up as so-called essential claims. It should be made clear what the client achieves by doing this. You should also evaluate whether you should support non-essential claims, which overlaps a bit with scope values.

connect/src/com/telenor/connect/utils/ClaimsParameterFormatter.java
```diff
@@ -9,6 +9,21 @@
...
+     *
+     * In other words, after successfully authorizing the request with the given Claims, the user
+     * will have data in the given fields.
+     *
+     * See http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter for more details.
+     *
+     * @param claims A claims object containing wanted claims.
+     * @return A JSON formatted string with all {@code claims} turned into essential claims.
+     * @throws JSONException
+     */
```
jorunfa: @ulfcarlin: Does this make sense?

ulfcarlin: yes

PR #33, opened 2016-06-13
Add getValidAccessToken method to ConnectSdk

jorunfa:
* AccessTokenCallback added. This is callback that gives you an Access Token as a string or an error.
* There are now `ConnectTokens` and `ConnectTokensTO`. The latter is the transfer object. The first can now tell you if the Access Token is expired or not.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectRefreshTokenMissingException.java
+++ b/connect/src/com/telenor/connect/ConnectRefreshTokenMissingException.java
@@ -1,0 +1,19 @@
+package com.telenor.connect;
+
+public class ConnectRefreshTokenMissingException extends IllegalStateException {
+    public ConnectRefreshTokenMissingException() {
+        super();
+    }
+
+    public ConnectRefreshTokenMissingException(String message) {
+        super(message);
+    }
+
+    public ConnectRefreshTokenMissingException(Throwable throwable) {
+        super(throwable);
+    }
+
+    public ConnectRefreshTokenMissingException(String message, Throwable throwable) {
+        super(message, throwable);
+    }
+}
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -9,6 +9,7 @@ import android.net.Uri;
 import android.text.TextUtils;
 
 import com.squareup.okhttp.HttpUrl;
+import com.telenor.connect.id.AccessTokenCallback;
 import com.telenor.connect.id.ConnectIdService;
 import com.telenor.connect.id.IdToken;
 import com.telenor.connect.id.TokenStore;
@@ -115,6 +116,20 @@ public final class ConnectSdk {
         Intent intent = getAuthIntent(parameters);
         intent.putExtra(ConnectUtils.CUSTOM_LOADING_SCREEN_EXTRA, customLoadingLayout);
         activity.startActivityForResult(intent, requestCode);
+    }
+
+    /**
+     * Get a valid Access Token. If a non-expired one is available, that will be given to the
+     * callback {@code onSuccess(String accessToken)} method.
+     *
+     * If it is expired, it will be refreshed and then returned. This requires a network call.
+     *
+     * @param callback callback that will be called on success or failure to update.
+     * @throws ConnectRefreshTokenMissingException if no Request Token is available
+     */
+    public static synchronized void getValidAccessToken(AccessTokenCallback callback) {
+        Validator.sdkInitialized();
+        sConnectIdService.getValidAccessToken(callback);
     }
 
     public static synchronized String getAccessToken() {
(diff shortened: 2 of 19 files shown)


connect/src/com/telenor/connect/id/TokenStore.java
```diff
@@ -39,7 +39,7 @@ public void update(ConnectTokens connectTokens) {
 
     public ConnectTokens get() {
         String connectTokensJson = context
-                .getSharedPreferences(PREFERENCES_FILE, Context.MODE_PRIVATE)
+                .getSharedPreferences(PREFERENCES_FILE,  Context.MODE_PRIVATE)
```
ulfcarlin: Double space => unnecessary change

connect/src/com/telenor/connect/ConnectRefreshTokenMissingException.java
```diff
@@ -0,0 +1,19 @@
+package com.telenor.connect;
+
+public class ConnectRefreshTokenMissingException extends IllegalStateException {
```
ulfcarlin: Are we not using checked exceptions? Is this an unrecoverable condition? Is this case really an illegal state?

connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -34,6 +34,19 @@ public ConnectIdService(
         this.tokenStore = tokenStore;
     }
 
+    public void getValidAccessToken(final AccessTokenCallback callback) {
+        if (retrieveTokens() == null) {
+            throw new ConnectRefreshTokenMissingException("Failed to call retrieveTokens().");
```
ulfcarlin: Is more than the refresh token missing? Could the message at least be improved?

connect/src/com/telenor/connect/ConnectRefreshTokenMissingException.java
jorunfa: Hmm, bit unsure here. If the user is signed in but we don't have (refresh) tokens, then something is very wrong with the state, I'd say.

connect/src/com/telenor/connect/id/ConnectIdService.java
jorunfa: Yes. Better now?

connect/src/com/telenor/connect/id/ConnectTokens.java
```diff
@@ -1,46 +1,38 @@
 package com.telenor.connect.id;
 
-import com.google.gson.annotations.SerializedName;
+import com.telenor.connect.utils.Validator;
 
-public class ConnectTokens {
-    @SerializedName("access_token")
-    private final String accessToken;
+import java.util.Calendar;
+import java.util.Date;
```
ulfcarlin: Consider using java.time on Java 8, or Joda time otherwise?

connect/src/com/telenor/connect/ConnectRefreshTokenMissingException.java
jorunfa: Extend simply from Exception instead?

connect/src/com/telenor/connect/id/IdToken.java
```diff
@@ -60,6 +64,10 @@ public String getSubject() {
         return subject;
     }
 
+    public Date getExpirationDate() {
```
ulfcarlin: Is this relevant to this commit? Was it mentioned in the commit message?

connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -36,7 +36,8 @@ public ConnectIdService(
 
     public void getValidAccessToken(final AccessTokenCallback callback) {
         if (retrieveTokens() == null) {
-            throw new ConnectRefreshTokenMissingException("Failed to call retrieveTokens().");
+            throw new ConnectRefreshTokenMissingException("retrieveTokens() returned null. Tokens" +
+                    "are missing.");
```
ulfcarlin: Missing space. Consider inserting the newline before the string literal instead, so that you do not need to split it up. I still think it is confusing to throw "ConnectRefreshTokenMissingException" when the client is trying to get an access token

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -117,6 +118,20 @@ public static synchronized void authenticate(Activity activity,
         activity.startActivityForResult(intent, requestCode);
     }
 
+    /**
+     * Get a valid Access Token. If a non-expired one is available, that will be given to the
+     * callback {@code onSuccess(String accessToken)} method.
```
ulfcarlin: Consider mentioning that the token may expire or become revoked before being used. If so, the client should try to refresh the tokens explicitly.

connect/src/com/telenor/connect/id/IdToken.java
jorunfa: Maybe not. I can remove it from this commit.

PR #43, opened 2016-09-14
Fix ID token expired bug

jorunfa:
If the ID token is expired the SDK will now get the current time from the internet, and see if it is indeed expired or not. This will stop the SDK from throwing an exception when the user has incorrect time/date settings.


connect/tests/src/test/java/com/telenor/connect/utils/InternetTimeTest.java
```diff
@@ -0,0 +1,29 @@
...
+public class InternetTimeTest {
+
+    @Test
+    public void internetTimeIsWithinBoundaryOfLocalTime() throws Exception {
+        final Date internetDate = InternetTime.getInternetDate();
+        final int boundarySeconds = 5;
+
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.SECOND, boundarySeconds);
+        Date upperLimit = calendar.getTime();
```
ulfcarlin: Consider adding final here and for lowerLimit and withinLimit, in order to be consistent about usage of the final keyword.

connect/src/com/telenor/connect/utils/InternetTime.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class InternetTime {
+
+    @SuppressLint("SimpleDateFormat")
+    private final static SimpleDateFormat DATE_FORMAT
+            = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz");
+
+    public static Date getInternetDate() throws IOException, ParseException {
+        final Request request = new Request.Builder()
+                .url("http://www.google.com")
```
ulfcarlin: Should we consider using one of our own servers/endpoints?

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import retrofit.client.Header;
+
+public class HeadersDateUtil {
+
+    @SuppressLint("SimpleDateFormat")
+    private final static SimpleDateFormat DATE_FORMAT
+            = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz");
+
+    @Nullable
+    public static Date get(@NonNull List<Header> headers) {
```
ulfcarlin: Consider naming this method something more descriptive, e.g. extractDate?

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
```diff
@@ -0,0 +1,39 @@
...
+        try {
+            return getDate(headers);
+        } catch (ParseException e) {
+            e.printStackTrace();
```
ulfcarlin: Is this appropriate exception handling?

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
```diff
@@ -0,0 +1,39 @@
...
+            return null;
+        }
+    }
+
+    @Nullable
+    private static Date getDate(@NonNull List<Header> headers) throws ParseException {
```
ulfcarlin: Consider combining this method into the method above

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
jorunfa: I believe the SDK shouldn't throw, at least. If the server should for some reason have a malformed Date header the client shouldn't crash. I don't know what else than printing the stack trace is appropriate. Thoughts?

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
jorunfa: Will do. Nested as `try -> for -> if`, or `for -> if -> try`?

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
jorunfa: Will fix.

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
ulfcarlin: I would prefer the latter, in order to reduce the scope of the try block as much as possible

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
ulfcarlin: I agree that the exception should not be propagated out. However: Is it normal to print stack traces to stderr in this repo specifically and in Android apps in general? Is there no application log? (Sorry about my ignorance)

jorunfa: Sorry about my ignorance; you are absolutely right. Just forgot about it for a second.

connect/src/com/telenor/connect/utils/HeadersDateUtil.java
jorunfa: Fixed.

connect/tests/src/test/java/com/telenor/connect/utils/IdTokenValidatorTest.java
```diff
@@ -146,7 +149,71 @@ public void missingIssueTimeThrows() throws Exception {
...
+    @Test
+    public void isValidExpirationTimeReturnsFalseOnNullExpDate() {
+        boolean actual = IdTokenValidator.isValidExpirationTime(null, new Date(), null);
+        assertThat(actual, is(false));
+    }
+
+    @Test
+    public void isValidExpirationTimeReturnsTrueWhenExpDateIsAfterCurrentDate() {
+        Calendar calendar = Calendar.getInstance();
+        Date localTime = calendar.getTime();
```
ulfcarlin: Consider naming this variable now to increase readability

connect/tests/src/test/java/com/telenor/connect/utils/IdTokenValidatorTest.java
```diff
@@ -146,7 +149,71 @@ public void missingIssueTimeThrows() throws Exception {
...
+
+        boolean actual = IdTokenValidator.isValidExpirationTime(future, localTime, null);
+        assertThat(actual, is(true));
+    }
+
+    @Test
+    public void isValidExpirationTimeReturnsFalseWhenCurrentTimeIsAfterExpAndServerTimestampIsMissing() {
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.HOUR, 1);
+        Date localTime = calendar.getTime();
```
ulfcarlin: Consider naming this variable "future" to increase readability. Consider setting up "now" and "future" in the same way as in the previous test

connect/tests/src/test/java/com/telenor/connect/utils/IdTokenValidatorTest.java
jorunfa: Better now?

PR #47, opened 2016-11-23
Add getAccessTokenExpirationTime and bump sdkVersion

jorunfa:


connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -58,6 +58,14 @@ public String getAccessToken() {
         return retrieveTokens().getAccessToken();
     }
 
+    public Date getAccessTokenExpirationTime() {
+        if (retrieveTokens() == null) {
```
stephan-nordnes-eriksen: Maybe cache result to variable so you don't call retrieveTokens twice?

jorunfa: I see what you mean. `retrieveTokens()` uses a cache and only actually fetches the tokens from the store if they are missing. It's a private method.

connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -58,6 +58,14 @@ public String getAccessToken() {
...
+            return null;
+        }
+
+        return retrieveTokens().getExpirationDate();
```
jorunfa: See above.

jorunfa: If I'm not misunderstanding something the `getString(…)` from `getSharedPreferences` also uses a cache internally, and blocks (is synchronous), so there shouldn't be race conditions here, I think?

connect/src/com/telenor/connect/id/ConnectIdService.java
stephan-nordnes-eriksen: Yes, but you should probably do something like
```
DataType datatype = retrieveTokens();
if(datatype == null){
//...

return datatype.getExpirationDate();
```

ps. I don't know what datatype should be used.

PR #50, opened 2017-05-02
Mc 1568 od

gorandostanic:
Add operator discovery functionality to Connect ID Android SDK allowing it to be used as MobileConnect SDK.

More info: https://confluence.skunk-works.no/jira/browse/MC-1568


connect/src/com/telenor/connect/utils/ConnectUrlHelper.java
```diff
@@ -60,6 +79,34 @@ public static Uri getAuthorizeUri(
         return builder.build();
     }
 
+    public static Uri getMobileConnectAuthorizeUri(
```
phyrex1an: This is almost identical to the getConnectAuthorizeUri. Can't you move the .appendPath(OAUTH_PATH).appendPath("authorize") into the ConnectSdkProfile and make them identical?

atanasdb: oh no :) I was typing that

connect/src/com/telenor/mobileconnect/id/MobileConnectAPI.java
```diff
@@ -0,0 +1,54 @@
...
+import retrofit.http.Header;
+import retrofit.http.Headers;
+import retrofit.http.POST;
+import retrofit.http.Path;
+
+public interface MobileConnectAPI {
+
+    @FormUrlEncoded
+    @Headers("Content-Type: application/x-www-form-urlencoded")
+    @POST("/oauth/{operator_prefix}/token")
```
atanasdb: That's a bit hardcoded.
I guess you need to use the endpoint paths from OD.

connect/src/com/telenor/connect/utils/ConnectUrlHelper.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/id/MobileConnectAPI.java
gorandostanic: Done!

atanasdb:
Could you make the WellKnownApi class available to ConnectId SDK profile as well, because we'll want to use the network_authentication_target_ips for the Force HE.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -409,8 +447,33 @@ public static IdToken getIdToken() {
...
     }
 
+    /**
+     * MobileConnect stuff
+     */
+
+    public static synchronized void sdkInitializeMobileConnect(
+            Context context,
+            OperatorDiscoveryConfig operatorDiscoveryConfig) {
+        if (!isInitialized()) {
```
atanasdb: Prefer using a guard statement

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -409,8 +447,33 @@ public static IdToken getIdToken() {
...
+                    fetchBooleanProperty(getApplicationInfo(context), USE_STAGING_PROPERTY),
+                    fetchBooleanProperty(getApplicationInfo(context), CONFIDENTIAL_CLIENT_PROPERTY));
+            context.registerReceiver(
+                    SimCardStateChangedBroadcastReceiver.getReceiver(),
+                    SimCardStateChangedBroadcastReceiver.getIntentFilter());
+        }
+    }
 
+    public static synchronized void sdkReinitializeMobileConnect() {
+        if (isInitialized()) {
```
atanasdb: Prefer using a guard statement

gorandostanic:
According to the book I am reading right now (Effective Java, remember? :-) it is always wise not to allow alien code to be executed in your synchronised block/method. Having read the rationale behind this, I couldn't agree more :)

I need an implementation of OperatorDiscoveryConfig interface to hold the defensive copy of the data. Hence the Plain-bla-bla thing.

ExampleOperatorDiscoveryConfig cannot serve that purpose because 1. It is in the example project outside the SDK 2. it is hard-coded 3. It is NOT our recommendation about how to implement that interface, just an example to make the example app work. 

Actually, I have written a disclaimer in README.md about it, which motivated me to make the defensive copy of the interface:

_The example implementation found in the example app is just a simple example. 
Developers are encouraged to implement their own scheme to keep this information secret._

atanasdb:
Yeah nice, but according to the same book:
"Classes containing methods or constructors whose invocation indicates a transfer of control cannot defend themselves against malicious clients. Such classes are acceptable only when there is mutual trust between the class and its client or when damage to the class’s invariants would harm no one but the client."

And that's what you have here. You do not have immutable class, that expose mutable objects through getters, and also messing up with the OperatorConfig would harm no one, but the client itself. 

Also having an immutable bean defined in the SDK, let's say OperatorDiscoveryConfig, amd instantiating it in the app, and passing it to the SDK, would have the same effect, and the code will be less.

gorandostanic:
Done! (applies to refactoring onAuthorize)

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -114,31 +114,20 @@ public Uri getAuthorizeUri(Map<String, String> parameters, List<String> locales)
     @Override
     public DoNext onAuthorize(Map<String, String> parameters) {
```
atanasdb: I think that this should be really onPreAuthorize , onPrepareAuthorize or something like that.
onAuthorize implies that the event happens after the authorization.

gorandostanic: Done! (newname = onStartAuthorization)

gorandostanic:
Regarding defensive copy: I'd like to keep it, if you agree (do you?). 

The amount of code added is not very big, but it still insulates our code from (potentially hard to debug) concurrency issues a developer might cause.

atanasdb:
But having a single immutable class (let's say implemented with immutables.org we use all over) will give you the same thing, no ?

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -224,9 +219,10 @@ private boolean initialize(OperatorDiscoveryAPI.OperatorDiscoveryResult odResult
                             getClientId(),
                             getRedirectUri()));
             isInitialized = true;
+            return true;
         } catch (RuntimeException ignored) {
         }
```
atanasdb: That's extremely fishy. Why do u need it?

atanasdb: (I mean the catch block, having problems with outlining the right line with GitHub :))

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    private OperatorDiscoveryAPI.OperatorDiscoveryResult getOperatorDiscoveryResult() {
+        TelephonyManager phMgr = (TelephonyManager) context.getSystemService(Context.TELEPHONY_SERVICE);
+        String networkOperator = phMgr.getNetworkOperator();
+        if (!TextUtils.isEmpty(networkOperator)) {
+            final String mcc = networkOperator.substring(0, 3);
+            final String mnc = networkOperator.substring(3);
+            Future<OperatorDiscoveryAPI.OperatorDiscoveryResult> operatorDiscoveryResultFuture =
+                    sExecutor.submit(new Callable<OperatorDiscoveryAPI.OperatorDiscoveryResult>() {
```
atanasdb: I'm not sure I understand this peace of code.
On the surface it looks like it is asynchronous, but actually it's quite synchronous isn't it?
Then why do we need the ExecutorService, and the Future if we could just make a synchronous retrofit call?
Also if you decide to refactor it to asynchronous, maybe you should use the facilities provided retrofit for making asynchronous calls, instead of implementing ones yourself.
I think a good example is in ConnectId SDK's part: getAccessTokenFromCode.

gorandostanic: I do not find it too fishy, though. Retrofit does basically the same thing behind the scenes in getAccessTokenFromCode, which is an async retrofit call: retrofit creates a parallel thread and gives you the requested object via the callback... well, after a while. 

I had reasons to want a synchronised retrofit call. This is indicated in retrofit by specifying a returning value rather than a callback parameter, but the thing is that one gets ```android.os.NetworkOnMainThreadException``` if this is to be executed in the main thread. It looks like synchronised retrofit calls are possible to execute only in parallel anyway, which is what I am doing here.

atanasdb: Fishy I meant about catching a RuntimeException and not handling it.

Concerning synchronous/async, I mean , why don't u just:
getOperatorDiscoveryApi().getOperatorDiscoveryResult_ForMccMnc(
                                    getOperatorDiscoveryAuthHeader(),
                                    operatorDiscoveryConfig.getOperatorDiscoveryRedirectUri(),
                                    mcc,
                                    mnc,
new Callback<ConnectTokensTO>() {
                    @Override
                    public void success(...Response response) {
                           odResult = response;
                           continue authorize.....
                    }

                            );

phyrex1an:
I agree with atanas. I don't get what the current OperatorDiscoveryConfig interface is supposed to accomplish, compared with making it a "struct".

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Done!

gorandostanic:
Scrapped the defensive copy thing, since you are all hostile to it :)

phyrex1an:
I wasn't hostile to the defensive copy, given the OperatorDiscoveryConfig interface. I just think that the interface is useless and should be an immutable object, and then you wont need the defensive copy.

Can you explain why OperatorDiscoveryConfig is an interface?

gorandostanic:
_Can you explain why OperatorDiscoveryConfig is an interface?_

1. Because it can be :)
2. To avoid any interfering in how developers should implement their own secret keeping schema
3. To "wash our hands" from the deficiency in OD we have talked about (and I believe we agreed upon), namely the need of storing secrets within the app. This is kind of documented in README.md

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -106,7 +105,7 @@ public static synchronized void authenticate(
             final Map<String, String> parameters,
             final int requestCode) {
         Validator.sdkInitialized();
-        sdkProfile.onStartAuthorization(parameters, new SdkProfile.OnStartAuthorizationCallback() {
+        sdkProfile.onStartAuthorization(parameters, new SdkProfile.OnStartAuthenticationCallback() {
```
atanasdb: I'm not sure about that renaming. Don't we want to use authorization everywhere instead of authentication.
@phyrex1an ?
However it's better that we are not consistent at least.

gorandostanic: Something here

phyrex1an:
You don't need an interface to accomplish 2 and 3. You just move the "defensive copy" out to the caller, how they store their secrets is up to them. The only difference is that their storage schema doesn't become mixed into an interface provided by us.

atanasdb:
You will need to declare your references modified by != threads with volatile.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -112,23 +115,74 @@ public Uri getAuthorizeUri(Map<String, String> parameters, List<String> locales)
...
+                        String networkOperator = phMgr.getNetworkOperator();
 
-        return DoNext.cancel;
+                        if (TextUtils.isEmpty(networkOperator)) {
+                            callback.onError();
+                            return;
+                        }
+
+                        if (!TextUtils.isEmpty(networkOperator)) {
+                        final String mcc = networkOperator.substring(0, 3);
```
atanasdb: Wrong indent

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -112,23 +115,74 @@ public Uri getAuthorizeUri(Map<String, String> parameters, List<String> locales)
...
+        String msisdn = readPhoneNumber();
+        if (msisdn != null) {
+            getOperatorDiscoveryApi().getOperatorDiscoveryResult_ForMsisdn(
+                    getOperatorDiscoveryAuthHeader(),
+                    new OperatorDiscoveryAPI.BodyForMsisdn(
+                            operatorDiscoveryConfig.getOperatorDiscoveryRedirectUri(),
+                            msisdn),
+                    odCallbackForMsisdn);
+        } else {
+            odCallbackForMsisdn.failure(null);
```
atanasdb: It's a bit misleading that you rely on calling a failure to handle the normal path. Could you refactor?

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -112,23 +115,74 @@ public Uri getAuthorizeUri(Map<String, String> parameters, List<String> locales)
...
 
-        final String msisdn = readPhoneNumber();
-        if (msisdn != null) {
-            odResult = getOperatorDiscoveryResult(msisdn);
-        }
+        final Callback<OperatorDiscoveryAPI.OperatorDiscoveryResult> odCallbackForMccMnc =
+                new Callback<OperatorDiscoveryAPI.OperatorDiscoveryResult>() {
+                    @Override
+                    public void success(OperatorDiscoveryAPI.OperatorDiscoveryResult operatorDiscoveryResult, Response response) {
+                        if (isInitialized || initialize(operatorDiscoveryResult)) {
```
atanasdb: Prefer
if (!isInitailized)
{ 
  initialize();
}

or just endsureInitialized() where you check the flag.

atanasdb: Also you actually care only about WellKnown stuff here, so maybe it'd be better if you check for it, rather than checking for initialization (and hide wellknown behind it).

connect/src/com/telenor/connect/ConnectSdk.java
phyrex1an: It's right that openid connect is also authentication, but since the flow is called "authorization code flow" I think we should just stick with authorization.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Since I can't use lambdas in this project, so I am not sure if it's worth it, it would be made only more unreadable. Besides: this can be arguably treated as a failure (i.e. a failure in getting the OD result for the phone's MSISDN since we haven't succeeded in reading the MSISDN).

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: 1. Done!
2. Well known is calculated in initialize() because it is supposed to be performed in CONNECT ID profile as well (not only MC).

gorandostanic:
I asked Mikael decide, he preferred builder :)

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
atanasdb: Yes, but I guess you could refactor the logic that happens in failure and put it in a mccMncFlow() method for example and then call this both from the onFailure and from the else branch.
Calling .failure to do your happy path is really hard to parse, for me at least.

gorandostanic: Done anyway!

connect/src/com/telenor/mobileconnect/operatordiscovery/OperatorDiscoveryConfig.java
```diff
@@ -30,30 +30,38 @@ public static Builder builder() {
         return new Builder();
     }
 
-    public static class Builder {
-        OperatorDiscoveryConfig operatorDiscoveryConfig;
+    public static final class Builder {
```
atanasdb: private fields are still there for the 

String operatorDiscoveryEndpoint;
etc.

gorandostanic: Done!

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -0,0 +1,96 @@
...
+
+    protected void deInitialize() {
+        isInitialized = false;
+        wellKnownConfig = null;
+    }
+
+    protected boolean initialize() {
+        if (!isInitialized) {
+            fetchWellKnownConfig();
+            isInitialized = true;
```
phyrex1an: This is not true as fetchWellKnownConfig will return before the wellKnownConfig variable has been set.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+    public void onStartAuthorization(
+            final Map<String, String> parameters,
+            final OnStartAuthorizationCallback callback) {
+
+        final Callback<OperatorDiscoveryAPI.OperatorDiscoveryResult> odCallbackForMccMnc =
+                new Callback<OperatorDiscoveryAPI.OperatorDiscoveryResult>() {
+                    @Override
+                    public void success(OperatorDiscoveryAPI.OperatorDiscoveryResult operatorDiscoveryResult, Response response) {
+                        if (!isInitialized()) {
+                            initialize(operatorDiscoveryResult);
```
phyrex1an: Likely we'd need a callback here.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+
+        private MobileConnectAPI mobileConnectApi;
+
+        MobileConnectAPIAdapter(MobileConnectAPI mobileConnectApi) {
+            this.mobileConnectApi = mobileConnectApi;
+        }
+
+        @Override
+        public void getAccessTokens(
+                String grant_type,
```
phyrex1an: camelCase?

connect/src/com/telenor/mobileconnect/SimCardStateChangedBroadcastReceiver.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    public static IntentFilter getIntentFilter() {
+        return SIM_CARD_CHANGED_INTENT_FILTER;
+    }
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        String state = intent.getExtras().getString("ss");
+        if (state != null && "READY".equals(state)) {
+            ConnectSdk.sdkReinitializeMobileConnect();
```
phyrex1an: What if this happens while an authentication is ongoing?

atanasdb:
> I think you should just say (and make sure in callbacks) that all sdk methods are to be called from the UI thread. That programming model seems saner than trying to mark the right fields volatile everywhere.
The reasoning behind the volatile reference is that, they are written in the UI thread, but in the next PR (forced_he) they'll be consumed in the WebView and thus in a thread launched by it.

gorandostanic:
_What happens if the sdk is de-initialized during a mobile connect authentication for example?_

Not much. Deinitialize() just flags the SDK to perform initialisation anew upon the next authenticate(). The current authentication (if any, since it is rather difficult to conceive one at this point of time) will be allowed to be performed with the old config.
 
_That programming model seems saner than trying to mark the right fields volatile everywhere._

Atanas reacted to this remark already.

_In many places (especially associated with initialize) you ignore the fact that some earlier method is not finished just because it returns, eg you have a callback but you try to go on as if the callback ran synchronously._

I think this is not wholly true. I believe I have made sure that the actual execution is always resumed from within the callback, it does not matter that it seemingly continues just like that. For example, initialize() is called only when a (sane) operatorDiscoveryResult is secured. If not, initialize() will never be called, and the activity will die. Anyway, I'd appreciate if you could be more specific here, I find Android development a bit weird, maybe I'm missing something substantial :)

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -409,8 +457,34 @@ public static IdToken getIdToken() {
...
+            OperatorDiscoveryConfig operatorDiscoveryConfig) {
+        if (isInitialized()) {
+            return;
+        }
+        sdkProfile = new MobileConnectSdkProfile(
+                context,
+                operatorDiscoveryConfig,
+                fetchBooleanProperty(getApplicationInfo(context), USE_STAGING_PROPERTY),
+                fetchBooleanProperty(getApplicationInfo(context), CONFIDENTIAL_CLIENT_PROPERTY));
+        context.registerReceiver(
```
jorunfa: Two things here:
1. With this setup you are always listening to sms being received, not just during a sign in process. I think unless this is needed, we should only listen to sms in the window where we actually must. So remember to unregister after you've gotten the SMS that you're looking for, and only listen for it when you are waiting for it.
2. `android.permission.READ_SMS` is a _dangerous_ permission, so you are not automatically granted it and must request it. It seems the code is registering a receiver before requesting this permission. And it seems it's only requesting it on button press, not if the user uses the `authenticate` method. That is not quite consistent with the existing code.

`ConnectWebViewClient` is a bit heavy class, but I would look at that to see how the asynchronous nature of sms is currently handled. The reason for the `Instruction` class is to notify Android to listen for SMS on a given page. If we don't currently have permission to read sms the instruction is stored, then permission is asked for, then the code reads the most recently received sms and starts listening for new ones.

connect/src/com/telenor/connect/ConnectSdkProfile.java
```diff
@@ -0,0 +1,114 @@
...
+    public ConnectSdkProfile(
+            Context context,
+            boolean useStaging,
+            boolean confidentialClient) {
+        super(context, useStaging, confidentialClient);
+    }
+
+    @Override
+    public HttpUrl getApiUrl() {
+        HttpUrl.Builder builder = new HttpUrl.Builder();
```
jorunfa: Maybe shorten/chain?
```
return new HttpUrl.Builder()
  .scheme()
  .host()
  .build()
```

connect/src/com/telenor/connect/ConnectSdkProfile.java
```diff
@@ -0,0 +1,114 @@
...
+        return null;
+    }
+
+    @Override
+    public String getRedirectUri() {
+        return redirectUri;
+    }
+
+    @Override
+    public String getExpectedIssuer(String actualIssuer) {
```
jorunfa: I'm confused about this method. I see it's overriding and thus must have the same signature, but I don't understand why it would ever need `actualIssuer`.

connect/src/com/telenor/connect/ConnectSdkProfile.java
```diff
@@ -0,0 +1,114 @@
...
+        if (getWellKnownConfig() != null) {
+            return getWellKnownConfig().getIssuer();
+        }
+        return getApiUrl() + OAUTH_PATH;
+    }
+
+    @Override
+    public List<String> getExpectedAudiences(List<String> actualAudiences) {
```
jorunfa: Same as previous, why would we need `actualAudiences` for a method called `getExpectedAudiences`? Should there be a method called `boolean audiencesAreExpected(List<String> actualAudiences)` or something like that?

connect/src/com/telenor/connect/utils/IdTokenValidator.java
```diff
@@ -27,31 +27,31 @@ public static void validate(final IdToken idToken, Date serverTimestamp) {
         }
 
         final String iss = idTokenClaimsSet.getIssuer();
```
jorunfa: Can this variable be renamed to `issuer`?

jorunfa: Or `idTokenIssuer`, or something more declarative?

connect/src/com/telenor/connect/utils/RestHelper.java
```diff
@@ -15,7 +18,35 @@
...
+
+    public static MobileConnectAPI getMobileConnectApi(String endpoint) {
+        return buildApi(endpoint).create(MobileConnectAPI.class);
+    }
+
+    public static WellKnownAPI getWellKnownApi(String endpoint) {
+        return buildApi(endpoint).create(WellKnownAPI.class);
+    }
+
+    public static OperatorDiscoveryAPI getOperatorDiscoveryAPI(String endpoint) {
```
jorunfa: `getOperatorDiscoveryAPI` -> `getOperatorDiscoveryApi`?

connect/src/com/telenor/mobileconnect/SimCardStateChangedBroadcastReceiver.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.telenor.connect.ConnectSdk;
+
+public class SimCardStateChangedBroadcastReceiver extends BroadcastReceiver {
+
+    private static SimCardStateChangedBroadcastReceiver sSimCardChangedReceiver =
+            new SimCardStateChangedBroadcastReceiver();
+
+    private static IntentFilter SIM_CARD_CHANGED_INTENT_FILTER =
+            new IntentFilter("android.intent.action.SIM_STATE_CHANGED");
```
jorunfa: Seems this intent isn't officially documented, so I would maybe add a comment, possibly linking to http://stackoverflow.com/questions/10528464/how-to-monitor-sim-state-change , for future reference.

connect/src/com/telenor/mobileconnect/ui/MobileConnectLoginButton.java
```diff
@@ -0,0 +1,32 @@
...
+        super(context, attributeSet);
+        setText(R.string.com_telenor_mobile_connect_login_button_text);
+        checkOrRequestReadPhoneStatePermission();
+    }
+
+    private void checkOrRequestReadPhoneStatePermission() {
+        Activity activity = getActivity();
+        if (activity.checkCallingOrSelfPermission(READ_PHONE_STATE) != PackageManager.PERMISSION_GRANTED) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                activity.requestPermissions(new String[]{READ_PHONE_STATE}, READ_PHONE_STATE_REQUEST_CODE);
```
jorunfa: 1. As mentioned earlier I don't think this is the place to ask for these permissions.
2. I don't see any implementation of the callback method `@Override onRequestPermissionsResult(...)` anywhere. You need to handle the user agree/decline.

mobileconnect-example/src/main/java/com/telenor/connect/mobileconnectexample/ExampleApplication.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        ConnectSdk.sdkInitializeMobileConnect(
+                getApplicationContext(),
+                OperatorDiscoveryConfig
+                        .builder()
+                        .endpoint("https://discover.mobileconnect.io/gsma/v2/discovery")
+                        .clientId("abce79a0-5a73-488e-9389-4bc7f2685b5b")
+                        .clientSecret("923901ca-6977-4f90-b18a-34da0c5c7b69")
```
jorunfa: FYI: This is an completely open repository, so this secret isn't secret anymore. I think this either has to be changed or commented on.

README.md
```diff
@@ -585,3 +585,54 @@ return the [Payment link](http://docs.telenordigital.com/apis/connect/payment/#s
 to your application. The Payment link is then used in a call to
 `ConnectSdk.initializePayment(Context, String)`, which opens a `ConnectActivity` where the user can
 perform the transaction.
+
+
+## Mobile Connect
```
gorandostanic: Done!

connect/res/values/strings.xml
```diff
@@ -5,4 +5,6 @@
     <string name="com_telenor_connect_error_text2">Please check your network settings or try again later.</string>
     <string name="com_telenor_connect_try_again">TRY AGAIN</string>
     <string name="com_telenor_connect_network_settings"><u>Network settings</u></string>
+    <string name="com_telenor_mobile_connect_login_button_text">Sign in with MobileConnect</string>
```
gorandostanic: In a way, it is, since I wanted them to be reviewed prior to having them translated. I'll request translations soon.

README.md
```diff
@@ -585,3 +585,54 @@ return the [Payment link](http://docs.telenordigital.com/apis/connect/payment/#s
...
+
+This SDK may be used for developing apps that authenticate users via MobileConnect. MobileConnect is a
```
gorandostanic: Done!

connect/res/values/strings.xml
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+        super(context, useStaging, confidentialClient);
+        this.operatorDiscoveryConfig = operatorDiscoveryConfig;
+    }
+
+    @Override
+    public HttpUrl getApiUrl() {
+        String host = operatorDiscoveryResult.getMobileConnectApiUrl().host();
+        if (useStaging) {
+            // will have no effect on non-Telenor subscribers
+            host = host.replace(
```
gorandostanic: I agree, but I found it handy during my manual tests, since the example app is registered with GSMA for production.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/SimCardStateChangedBroadcastReceiver.java
gorandostanic: Not much, since this is just a flag. The current auth. flow will be allowed to continue with the old config. A new auth. flow will reinitialize itself.

mobileconnect-example/src/main/java/com/telenor/connect/mobileconnectexample/ExampleApplication.java
gorandostanic: It is documented, there is a disclaimer in README.md. 

The problem is in the nature of Operator Discovery. We (the MC team) have discussed this issue during our standups, and agreed on making a simplest possible example which "hides" its secrets in the compiled code, simultaneously encouraging developers to implement their own secret hiding scheme. The fact that we don't know exactly how they are supposed to do that, it is another issue. That's why Operator Discover sucks. The architecture of it is a political decision, rather than a technical one :(

connect/src/com/telenor/connect/ConnectSdkProfile.java
```diff
@@ -0,0 +1,114 @@
...
+                .buildUpon()
+                .appendPath(OAUTH_PATH)
+                .appendPath("authorize")
+                .build();
+    }
+
+    @Override
+    public void onStartAuthorization(Map<String, String> parameters, OnStartAuthorizationCallback callback) {
+        initialize();
+        callback.onSuccess();
```
gorandostanic: Done! I renamed the method and changed its signature to initializeAndContinueAuthorizationFlow(callback), to reflect what it's doing better.

connect/src/com/telenor/connect/AbstractSdkProfile.java
gorandostanic: Done! (see previous comment)

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -0,0 +1,96 @@
...
+        }
```
gorandostanic: This is executed only when SdkProfile.authenticate() is called, which is synchronised.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    @Override
+    public String getRedirectUri() {
+        return operatorDiscoveryConfig.getOperatorDiscoveryRedirectUri();
+    }
+
+    @Override
+    public String getExpectedIssuer(String actualIssuer) {
+        // discrepancy between .well-known configuration and the actual issuer returned
```
gorandostanic: This is a workaround, the code needs changing as soon as this discrepancy is fixed.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+        }
+        if (getWellKnownConfig() != null) {
+            return getWellKnownConfig().getIssuer();
+        }
+        return actualIssuer;
+    }
+
+    @Override
+    public List<String> getExpectedAudiences(List<String> actualAudiences) {
+        return new ArrayList<>(actualAudiences);
```
gorandostanic: For CONNECT ID, it works just like you say. For Mobile Connect, however, I have another discrepancy here.... our own server returns `{operator};{clientId}` (like "telenor-norway;weoirasdhfkjsd"). I have checked GSMA documentation, and I haven't found a recommendation what expected audience should be. Hence the code.

mobileconnect-example/src/main/java/com/telenor/connect/mobileconnectexample/ExampleApplication.java
atanasdb: Hiding secrets in code (obfuscation)  is different than hiding them from a public repo.
AFAIK we're fine with sharing this secret here, as the client has been registered explicitly for being used as a sample.

connect/src/com/telenor/mobileconnect/ui/MobileConnectLoginButton.java
gorandostanic: Done!

Actually, I scrapped reading phone numbers entirely because:
1. It is hard to find the right place to ask for READ_PHONE_STATE permission without exposing developers to additional activities in the manifest.
2. It complicates the code a lot.
3. Phone reading works only if the phone number is stored on the SIM card, which is not the case with the vast majority of phones anyway. For example, it doesn't work on my Yotaphone.

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void setConnectIdService(ConnectIdService connectIdService) {
+        this.connectIdService = connectIdService;
+    }
+
+    protected boolean isInitialized() {
+        return isInitialized;
+    }
+
+    protected void  setInitialized(boolean isInitialized) {
```
gorandostanic: Done!

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -0,0 +1,96 @@
...
+        this.isInitialized = isInitialized;
+    }
+
+    protected void deInitialize() {
+        isInitialized = false;
+        wellKnownConfig = null;
+    }
+
+    protected boolean initialize() {
```
gorandostanic: Done! (refactored)

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: Done!

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: This is just (a temporary) workaround around a bug in comoyoauth impacting MobileConnect. Connect ID profile ignores this parameter, but MobileConnect profile doesn't.

I created a Jira task for that (https://confluence.skunk-works.no/jira/browse/MC-1612), we'll fix it in the next sprint, I think.

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: Almost the same as the actualIssuer thing (https://confluence.skunk-works.no/jira/browse/MC-1613)

connect/src/com/telenor/connect/utils/IdTokenValidator.java
gorandostanic: Done (renamed to issuer)

connect/src/com/telenor/connect/utils/RestHelper.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -0,0 +1,316 @@
...
+        return operatorDiscoveryResult.getClientSecret();
+    }
+
+    @Override
+    public String getRedirectUri() {
+        return operatorDiscoveryConfig.getOperatorDiscoveryRedirectUri();
+    }
+
+    @Override
+    public String getExpectedIssuer(String actualIssuer) {
```
gorandostanic: This is just (a temporary) workaround around a bug in comoyoauth impacting MobileConnect. Connect ID profile ignores this parameter, but MobileConnect profile doesn't.

I created a Jira task for that (https://confluence.skunk-works.no/jira/browse/MC-1612), we'll fix it in the next sprint, I think.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/SimCardStateChangedBroadcastReceiver.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
atanasdb: Even after the fixes, I agree with @jorunfa that the naming here is not as good as it could be.

connect/src/com/telenor/connect/ConnectSdkProfile.java
atanasdb: Even after the fixes, I agree with @jorunfa that the naming here is not as good as it could be.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -184,17 +133,17 @@ private void tryMccMnc(
                 operatorDiscoveryConfig.getOperatorDiscoveryRedirectUri(),
                 mcc,
                 mnc,
-                odCallbackForMccMnc);
-    }
+                new Callback<OperatorDiscoveryAPI.OperatorDiscoveryResult>() {
```
atanasdb: Great that u did that with the anonymous callbacks :)
I wanted to write that a few times and I always was interrupted by something.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: I'll scrap actualIssuer argument as soon as the fixes are in place, so the signature's gonna be getExpectedIssuer() only.

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: I'll scrap actualIssuer argument as soon as the fixes are in place, so the signature's gonna be getExpectedIssuer() only.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: They were actually anonymous all the time (in the sense "anonymous class"), but I needed references in order to reuse them, like passing them as an argument to another method (okay, a could have done with odCallbackForMccMnc only, but I created odCallbackForMsisdn for the sake of clarity) . 

Having scrapped the msisdn-based operator discovery, I do not have to refer explicitely to callbacks any more. Hence the change.

atanasdb: Yeah I meant referencing them.
But anyway, now that you have also changed the .well-known fetching to happen first before proceeding to the authorize (it's not in parallel anymore), I think it'll be better to construct some WellKnown immutable struct and pass it to the Authorization/WebView. That way you won't rely on the volatile references and the model is easier to understand I believe.

gorandostanic: Yeah, it is good idea. I'll see what I can do.

gorandostanic:
Now I’ve fixed the old unit tests and added some more for MC :-)

I think the things are more or less in place now... apart from the following:

1. The expected/actual/issuer/audience thing. 
I’ll do the fix it as soon as the related bug in comoyoauth has been fixed. 

2. The translations (i.e. two untranslated strings used exclusively in Mobile Connect). 
We (i.e. the MC gang :) have discussed this issue recently and have come to a conclusion that we'd rather have this PR approved without translations this time, as being a very new thing. In addition, translations may take some time before they are ready to pick.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -113,13 +129,31 @@ private static Intent getAuthIntent(Map<String, String> parameters) {
...
+            public void onError() {
+                showAuthCancelMessage(activity);
+            }
+        });
+    }
+
+    private static void showAuthCancelMessage(Activity activity) {
+        Toast.makeText(
+                activity,
+                R.string.com_telenor_authorization_cancelled,
```
gorandostanic: MC only, will never reach that code if Connect Id

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -145,20 +179,23 @@ public static Fragment getAuthFragment(Map<String, String> parameters) {
...
     public static synchronized void getValidAccessToken(AccessTokenCallback callback) {
         Validator.sdkInitialized();
-        sConnectIdService.getValidAccessToken(callback);
+        sdkProfile.getConnectIdService().getValidAccessToken(callback);
     }
 
     public static synchronized String getAccessToken() {
         Validator.sdkInitialized();
-        return sConnectIdService.getAccessToken();
+        if (sdkProfile.getConnectIdService() != null) {
```
gorandostanic: MC is doing lazy initialisation (not before the first authenticate request), so this is necessary in case the app is trying to get tokens without starting the auth flow first. Does not apply to Connect Id, though, since it is initialised statically.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -373,6 +393,34 @@ private static void loadConnectConfig(Context context) {
             String paymentSuccessUriString = (String) paymentSuccessUriObject;
             sPaymentSuccessUri = paymentSuccessUriString;
         }
+        return profile;
+    }
+
+    private static ApplicationInfo getApplicationInfo(Context context) {
+        ApplicationInfo ai = null;
```

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -373,6 +393,34 @@ private static void loadConnectConfig(Context context) {
...
+                    context.getPackageName(), PackageManager.GET_META_DATA);
+        } catch (PackageManager.NameNotFoundException e) {
+            return null;
+        }
+        return ai;
+    }
+
+
+    private static boolean fetchBooleanProperty(ApplicationInfo appInfo, String propertyName) {
+        if (appInfo != null) {
```

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -409,8 +457,34 @@ public static IdToken getIdToken() {
      */
     public static void getUserInfo(Callback<UserInfo> userInfoCallback) {
         Validator.sdkInitialized();
-        sConnectIdService.getUserInfo(userInfoCallback);
+        sdkProfile.getConnectIdService().getUserInfo(userInfoCallback);
     }
 
+    /**
+     * MobileConnect stuff
```
gorandostanic: No (removed!)

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
gorandostanic: Since the fix in comoyoauth is in code review now, I removed this workaround

(see https://gerrit.c.skunk-works.no/#/c/45471/1)

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: Since the fix in comoyoauth is in code review now, I removed this workaround

(see https://gerrit.c.skunk-works.no/#/c/45471/1)

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: Since the fix in comoyoauth is in code review now, I removed this workaround

(see https://gerrit.c.skunk-works.no/#/c/45471/1)

connect/src/com/telenor/connect/ConnectSdk.java
jorunfa: Ah, sorry, might have misunderstood that one a bit.
The dev can currently choose which context to use, so it might be an application context or it might be an activity context. In the latter case I would expect it to die if the activity dies.

connect/src/com/telenor/mobileconnect/ui/MobileConnectLoginButton.java
```diff
@@ -0,0 +1,20 @@
...
+import android.app.Activity;
+import android.content.Context;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.util.AttributeSet;
+
+import com.telenor.connect.R;
+import com.telenor.connect.ui.ConnectLoginButton;
+
+import static android.Manifest.permission.READ_PHONE_STATE;
```
jorunfa: Is this import needed now?

connect/tests/src/test/java/com/telenor/TestHelper.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final long SLEEPING_TIME_IN_MILLIES = 10;
+
+    public interface BooleanSupplier {
+        boolean getAsBoolean();
+    }
+
+    public static boolean flushForegroundTasksUntilCallerIsSatisifed(
+            long maxMillies,
+            BooleanSupplier isCallerSatisfied) {
+        for (; maxMillies > 0; maxMillies -= SLEEPING_TIME_IN_MILLIES) {
```
jorunfa: Shouldn't this be a while loop?

connect/tests/src/test/java/com/telenor/connect/ui/ConnectLoginButtonTest.java
```diff
@@ -28,28 +36,41 @@
...
+        WELL_KNOWN_API_MAP.put(
+                ((ConnectSdkProfile) ConnectSdk.getSdkProfile()).getWellKnownEndpoint(),
+                MOCKED_VALID_WELL_KNOWN_API);
+
+        final Activity activity = Robolectric.buildActivity(TestActivity.class).create().get();
+        final ConnectLoginButton button = (ConnectLoginButton) activity.findViewById(R.id.login_button);
         button.setLoginScopeTokens("profile");
         button.performClick();
 
+        assertThat("Is activity started?",
```
jorunfa: `assertThat("Is activity started?",` ->
`assertThat("Activity is started",`
would make more sense, I think. Reads like _assert that activity is started_.

connect/tests/src/test/java/com/telenor/mobileconnect/MobileConnectTestHelper.java
```diff
@@ -0,0 +1,105 @@
...
+     * when tests are to be executed in Android Studio w/Gradle.
+     *
+     * Should have no effect in all other cases.
+     *
+     * See: https://github.com/robolectric/robolectric/issues/1592
+     */
+    public static class CustomRobolectricTestRunner extends RobolectricTestRunner {
+        public CustomRobolectricTestRunner(Class<?> testClass) throws InitializationError {
+            super(testClass);
+            if (System.getProperty("android.resources") == null) {
```
jorunfa: Avoid arrow anti-pattern, reverse the if's and make them guards?
```java
if (System.getProperty("android.resources") != null) {
  return;
}
String pathToMerged = // ...
```

mobileconnect-example/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,34 @@
...
+        android:name="com.telenor.connect.mobileconnectexample.ExampleApplication"
+        android:supportsRtl="true"
+        android:theme="@style/AppTheme" >
+        <activity android:name="com.telenor.connect.mobileconnectexample.SignedInActivity" />
+        <activity android:name="com.telenor.connect.mobileconnectexample.SignInActivity">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>>
```
jorunfa: Remove extra `>`

mobileconnect-example/src/main/java/com/telenor/connect/mobileconnectexample/ExampleApplication.java
jorunfa: Okay. Just a heads up, my experience is that third party devs tend to struggle when I have trouble with the bits they are supposed to implement.

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -0,0 +1,95 @@
...
+    protected void setInitialized(boolean isInitialized) {
+        this.isInitialized = isInitialized;
+    }
+
+    protected void deInitialize() {
+        isInitialized = false;
+        wellKnownConfig = null;
+    }
+
+    protected void initializeAndContinueAuthorizationFlow(final OnStartAuthorizationCallback callback) {
```
gorandostanic: Not sure about this, it is a class. Did you mean callback -> onStartAuthorizationCallback?

connect/tests/src/test/java/com/telenor/TestHelper.java
gorandostanic: Done

connect/tests/src/test/java/com/telenor/connect/ui/ConnectLoginButtonTest.java
gorandostanic: Done!

connect/tests/src/test/java/com/telenor/mobileconnect/MobileConnectTestHelper.java
gorandostanic: Done!

connect/src/com/telenor/mobileconnect/ui/MobileConnectLoginButton.java
gorandostanic: Done!

mobileconnect-example/src/main/AndroidManifest.xml
gorandostanic: Done!

atanasdb:
You have the "hack" for the staging environment of Telenor for the authorize host, which is fine, but you do not have the same behaviour for the .well-known uri construction which leads to problems with the forced he. Could you fix that?

gorandostanic:
Sorry for that, I just happened to forget it, maybe because it shouldn't be
used for MC, since the environment is decided upon client registration.
I'll fix it as soon as I am on the plane :)

G.

atanasdb:
Yes, best if you don't go over that url rewrite for MC at all.

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -122,8 +122,8 @@ public void onStartAuthorization(final OnStartAuthorizationCallback callback) {
             return;
         }
 
-        final String mcc = networkOperator.substring(0, 3);
-        final String mnc = networkOperator.substring(3);
+        final String mcc = "242"; //networkOperator.substring(0, 3);
```
atanasdb: oops

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -92,4 +92,11 @@ public void failure(RetrofitError error) {
                     }
                 });
     }
+
+    protected String applyStagingOnEndpoint(String endpoint) {
```
atanasdb: I don't think this has to be a member of the base class.
It's kind of a hack, so we'd better have it as limited as possible.
Plus, in the future (when you are ready with the tests), you want that removed for the MC profile, and then it make no sense to have it here. This is purely ConnectId specific thing.

gorandostanic:
Connect ID: tested manually in accord with the test scenarios on:

SA Galaxy Trend 2 (SM-G313HN, Android 4.4.2)
Yotaphone 2 (YD201, Android 5.0)
SA Galaxy S5 (SM-GF00F, Android 6.0.1)

Wasn't quiet sure what 6. meant (Denying app permissions after authentication -> close sign in flow :-), so I skipped that one.

Regarding Mobile Connect: could not test in Belgrade fully because of a roaming issue (mcc/mnc), but performed the tests applicable to the current state of Mobile Connect SDK with some cheating.

PR #51, opened 2017-05-03
Allow HE to be tried based on mcc/mnc information

atanasdb:
Jira: MC-1556

HE service is being extended to support Force HE over WiFi. For the
purpose we need to support HE flows started based on the MCC/MNC
passed by the device to the authorization endpoint.

See design doc at
https://confluence.skunk-works.no/confluence/x/30BRAg.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -8,7 +8,9 @@ import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.telephony.TelephonyManager;
 import android.text.TextUtils;
+import android.webkit.CookieManager;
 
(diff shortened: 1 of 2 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -306,6 +308,19 @@ public static synchronized void sdkInitialize(Context context) {
                 sRedirectUri);
     }
 
+    public static String getMccMncLoginHint() {
```
gorandostanic: Consider splitting this method into getMccMnc() and getMccMncLoginHint(String mccMnc), so we can use both.

atanasdb: Done. I was experimenting a bit with the workflow here and just amended the last commit (which is not you should do in GitHub :) ) , so you won't be able to see the prev commit and the diff.

PR #53, opened 2017-05-29
Forced HE over WiFi

atanasdb:
Add the possibility to execute request towards specific IPs found from
the .well-known configuration API using the cellular network interface, thus
achieving successful HE ,even when both WiFi and cellular are on.

For more information, see design doc at:
https://confluence.skunk-works.no/confluence/x/30BRAg.


atanasdb:
Original PR in https://github.com/gorandostanic/connect-android-sdk/pull/1

PR #54, opened 2017-05-30
Add operator selection to Mobile Connect authentication flow

gorandostanic:
In order to enable Mobile Connect authentication on devices without SIM card, we divert auth flow to operator selection. The actual initialisation is delayed until the interception of the redirect URI.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -160,9 +160,6 @@ public final class ConnectSdk {
                             Uri startUri,
                             boolean isDivertedToFetchInitUri) {
 
-                        if (TextUtils.isEmpty(parameters.get("state"))) {
-                            parameters.put("state", UUID.randomUUID().toString());
-                        }
                         sLastAuthState = parameters.get("state");
 
                         Intent intent = getAuthIntent(startUri);
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -86,6 +86,8 @@ public final class ConnectSdk {
      */
     public static final String USE_STAGING_PROPERTY = "com.telenor.connect.USE_STAGING";
 
+    public static final int NO_CUSTOM_LAYOUT = -1;
+
     public static final String ACTION_LOGIN_STATE_CHANGED =
             "com.telenor.connect.ACTION_LOGIN_STATE_CHANGED";
 
@@ -116,27 +118,14 @@ public final class ConnectSdk {
             final Activity activity,
             final Map<String, String> parameters,
             final int requestCode) {
-        Validator.sdkInitialized();
-        sdkProfile.onStartAuthorization(new SdkProfile.OnStartAuthorizationCallback() {
-            @Override
-            public void onSuccess() {
-                Intent intent = getAuthIntent(parameters);
-                activity.startActivityForResult(intent, requestCode);
-            }
-
-            @Override
-            public void onError() {
-                showAuthCancelMessage(activity);
-            }
-        });
-
+        authenticate(activity, parameters, NO_CUSTOM_LAYOUT, requestCode);
     }
 
-    private static Intent getAuthIntent(Map<String, String> parameters) {
+    private static Intent getAuthIntent(Uri startUri) {
         final Intent intent = new Intent();
         intent.setClass(getContext(), ConnectActivity.class);
         intent.setAction(ConnectUtils.LOGIN_ACTION);
-        final String url = getAuthorizeUriAndSetLastAuthState(parameters).toString();
+        final String url = startUri.toString();
         intent.putExtra(ConnectUtils.LOGIN_AUTH_URI, url);
         return intent;
     }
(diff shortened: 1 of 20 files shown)


connect/src/com/telenor/connect/SdkProfile.java
```diff
@@ -24,10 +24,23 @@
...
+            Map<String, String> parameters,
+            List<String> uiLocales,
+            OnStartAuthorizationCallback callback);
+
+    interface OnDeliverAuthorizationUriCallback {
+        void onSuccess(Uri authorizationStartUri);
         void onError();
     }
+
+    interface OnStartAuthorizationCallback extends OnDeliverAuthorizationUriCallback {
```
phyrex1an: Any reason why having this is the same hierarchy? Why isn't there a completely separate callback for mobile connect for the operator discovery result (irregardless of the result was delivered via the manual selection or automatically)? It looks weird to have both kinds of callbacks in the same class (eg, sometimes you implement the onDivert method by throwing RuntimeException).

connect/src/com/telenor/mobileconnect/MobileConnectSdkProfile.java
```diff
@@ -100,32 +103,123 @@ public Uri getAuthorizeUri(Map<String, String> parameters, List<String> locales)
     }
 
     @Override
-    public void onStartAuthorization(final OnStartAuthorizationCallback callback) {
+    public synchronized void onStartAuthorization(
```
phyrex1an: I really dislike these synchronized methods, including the already present ones. The scope is way to wide, and it's not at all clear what is the actions which needs serialization. Can you rewrite into using explicit locks instead, with as small scope as possible? Alternatively get rid of the volatile usage in favor of explicit state passing between threads, but I think we already discussed that and said it was infeasible in these patches...

connect/src/com/telenor/connect/ui/ConnectWebViewClient.java
```diff
@@ -87,11 +88,47 @@ public ConnectWebViewClient(
     }
 
     @Override
-    public boolean shouldOverrideUrlLoading(WebView view, String url) {
-        if (ConnectSdk.getRedirectUri() != null
-                && url.startsWith(ConnectSdk.getRedirectUri())) {
-            ConnectUtils.parseAuthCode(url, connectCallback);
-            return true;
+    @SuppressWarnings("unchecked")
+    public boolean shouldOverrideUrlLoading(final WebView view, final String url) {
```
phyrex1an: This method looks quite misnamed now.

connect/src/com/telenor/connect/SdkProfile.java
atanasdb: I agree with that.
Also for me Operator Selection should be just a different activity, not part of the ConnectActivity.
And you should not modify ConnectWebViewClient.java with the code for Operator Selection.
The flow should be:
getMccMnc -> yes -> getWellKnown -> startAuthroization(wellKnown)
getMccMnc -> no -> getOSUri -> getOperator -> getWellKnown -> startAuthroization(wellKnown)

atanasdb: Why does my comment goes after Goran's even though ts are correct :)?
"I agree with that" is to the @phyrex1an's comment.

connect/src/com/telenor/connect/ui/ConnectWebViewClient.java
atanasdb: I think, the name confusion comes form the fact you are placing code where it does not really belong. See me other comment.

connect/src/com/telenor/connect/SdkProfile.java
gorandostanic: I kind of disagree for a couple of reasons listed below (ideally, operator selection should have been a part of a normal OAuth2 flow, in which case we wouldn't have had discussions of this kind in the first place). 

1. A separate OperatorSelection activity would require publishing it in the manifest, thus exposing developers to an activity they are not supposed to use. I think they generally benefit from encapsulations of this kind (the less - the better).

2. ConnectWebViewClient is not modified with OperatorSelection code, but with a more generic thing that the MobileConnect uses for operator selection. Connect Id could have used it for something else, provided they decided to implement preamble diversions. The fact that they don't do it now does not mean they couldn't.

atanasdb: 1. I'm not sure I understand this argument. Putting an activity in a manifest, shares this with the OS, and does not expose developers to it more than it is currently. Also the encapsulation you did, makes the code a bit harder to follow.
2. "Connect Id could have used it for something else" like what? I think this shouldOverrideUrlLoading is the wrong hook to achieve what you want.

gorandostanic:
1. What I meant was that it's developers that publish activities in the
manifest, so I think it's generally detrimental that they have to publish
activities they are not supposed to use themselves, merely to keep the SDK
in shape.
2. I am not sure it's a wrong hook... my understanding is that the hook is
there to answer "yes" if you don't want the url to be rendered by the
WebView, giving the application a chance to do something else instead.

connect/src/com/telenor/connect/SdkProfile.java
gorandostanic: 1. What I meant was that it's developers that publish activities in the
manifest, so I think it's generally detrimental that they have to publish
activities they are not supposed to use themselves, merely to keep the SDK
in shape.
2. I am not sure it's a wrong hook... my understanding is that the hook is
there to answer "yes" if you don't want the url to be rendered by the
WebView, giving the application a chance to do something else instead.

atanasdb: 1. They don't have to publish anything if you move the activity into the SDK manifest.

2.  You are putting something that's very generally used in a very low level implementation hook. E.g. let's say they don't use WebViews, but something else? Do you still want your OperatorSelection handling to live there.
2b smells like YAGNI, KISS pls :)

atanasdb: And "something else" from 2. is quite real: they have plans to drop the WebView.

gorandostanic: OK, you have convinced me, I'll do a rework :)

atanasdb: yay

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -146,22 +135,59 @@ public static synchronized void authenticate(final Activity activity,
...
-            public void onError() {
-                showAuthCancelMessage(activity);
-            }
+                    @Override
+                    public void onSuccess(Uri authorizationStartUri) {
+                        startAuthActivity(authorizationStartUri, false);
+                    }
+
+                    @Override
+                    public void onDivert(Uri diversionUri) {
```
phyrex1an: I'm guessing the work done per https://github.com/telenordigital/connect-android-sdk/pull/54/files/8514c618036b23ada24c6ba55e45b03c3ad6c0b2#diff-edfffb4e70b7f3f2268239f7a63b2636R43 moves this method somewhere else? If that's the case, consider if the naming could be made less generic.

PR #55, opened 2017-06-02
Add sending of wellKnownConfig along with the intent

gorandostanic:


connect/src/com/telenor/connect/WellKnownAPI.java
```diff
@@ -15,7 +16,7 @@
     @GET("/")
     void getWellKnownConfig(Callback<WellKnownConfig> callback);
 
-    class WellKnownConfig {
+    class WellKnownConfig implements Serializable {
```
phyrex1an: My (lacking) understanding is that Parcelable is prefered over Serializable.

gorandostanic: Done! (implemented Parceable instead)

connect/src/com/telenor/connect/WellKnownAPI.java
```diff
@@ -28,5 +33,44 @@ public String getIssuer() {
...
+
+        @Override
+        public int describeContents() {
+            return 0;
+        }
+
+        @Override
+        public void writeToParcel(Parcel dest, int flags) {
+            dest.writeString(issuer);
+            if (networkAuthenticationTargetIps == null) {
```
phyrex1an: So this doesn't make a distinction between a null networkAuthenticationTargetIps and an empty networkAuthenticationTargetIps so null is always deserialized to empty. Rest of the code seems to treat both those cases the same so it probably doesn't matter. If it's easy you can try making networkAuthenticationTargetIps always non-null so that case doesn't need to be handled.

connect/src/com/telenor/connect/WellKnownAPI.java
```diff
@@ -28,5 +33,44 @@ public String getIssuer() {
         public Set<String> getNetworkAuthenticationTargetIps() {
             return networkAuthenticationTargetIps;
         }
+
+        protected WellKnownConfig(Parcel in) {
+            issuer = in.readString();
+            networkAuthenticationTargetIps = new HashSet<>();
+            int count = in.readInt();
```
phyrex1an: Mini tip: If you switch these two lines you can initialize the hash set to the right size. No benefit for small sets though.

connect/src/com/telenor/connect/WellKnownAPI.java
gorandostanic: Done!

I mean, it's a bit tricky to make sure that the variable networkAuthenticationTargetIps is always non-null (I think null value will be injected if the SeralizedName is missing in JSON), but it's much easier to achieve the same with the getter (which I did :-)

connect/src/com/telenor/connect/WellKnownAPI.java
gorandostanic: Done!

PR #60, opened 2017-06-23
Make the Mobile Connect part of this SDK more Mobile Connect compliant

gorandostanic:


connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -3,14 +3,21 @@
...
+import java.util.Map;
+import java.util.UUID;
 
 import retrofit.Callback;
 import retrofit.RetrofitError;
 import retrofit.client.Response;
 
 public abstract class AbstractSdkProfile implements SdkProfile {
-
+    private final String STATE = "state";
```
atanasdb: I think the repo style is to use the string literal and do not defined constants and you should adhere to that.

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -73,6 +82,16 @@ protected void deInitialize() {
         wellKnownConfig = null;
     }
 
+    @Override
+    public void onStartAuthorization(
```
atanasdb: TextUtils.isEmpty

connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -96,4 +115,21 @@ public void failure(RetrofitError error) {
...
     }
+
+    @Override
+    public String getLastAuthState() {
+        return lastAuthState;
+    }
+
+    @Override
+    public void validateTokens(ConnectTokensTO tokens, Date serverTimestamp) {
+        Validator.notNullOrEmpty(tokens.getAccessToken(), "access_token");
```
atanasdb: I think you are missing fix for one of the issue's requirements:
Do not fail of scope is not present in the /token response.

gorandostanic: Done!

atanasdb: hmm, are u sure, did u test it with an empty scope?
doesn't json parser expects scope when parsing ConnectTokensTO ?

connect/src/com/telenor/connect/ConnectSdkProfile.java
```diff
@@ -124,4 +128,12 @@ private String applyUseStaginfOnEndpoint(String endpoint) {
         }
         return endpoint.replace("connect.telenordigital.com", "connect.staging.telenordigital.com");
     }
+
+    @Override
+    public void validateTokens(ConnectTokensTO tokens, Date serverTimestamp) {
+        super.validateTokens(tokens, serverTimestamp);
+        Validator.notNullOrEmpty(tokens.getScope(), "scope");
+        Validator.notNull(tokens.getExpiresIn(), "expires_in");
+        Validator.notNullOrEmpty(tokens.getRefreshToken(), "refresh_token");
```
phyrex1an: Connect will return a response without refresh token in case there is something wrong with the db, the access token will still be valid in that case.

connect/src/com/telenor/connect/AbstractSdkProfile.java
phyrex1an: ^ Unanswered comment above.

gorandostanic: Hehe, we settled it in spoken :) But yeah, sure, i tested the parser, it worked.

connect/src/com/telenor/connect/ConnectSdkProfile.java
gorandostanic: This check has always been in the SDK.  If it has to be corrected, I'd like to fix it in the context of a separate Jira issue.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -136,18 +135,33 @@ private static Intent getAuthIntent(Map<String, String> parameters) {
...
+            parameters.put("login_hint", String.format("MCCMNC:%s", mccMnc));
+        }
         final String url = getAuthorizeUriAndSetLastAuthState(parameters).toString();
         intent.putExtra(ConnectUtils.LOGIN_AUTH_URI, url);
-        intent.putExtra(ConnectUtils.WELL_KNOWN_CONFIG_EXTRA, sdkProfile.getWellKnownConfig());
+        intent.putExtra(ConnectUtils.WELL_KNOWN_CONFIG_EXTRA, wellKnownConfig);
         return intent;
     }
 
+    private static boolean isEmpty(Collection c) {
```
atanasdb: I'm not sure this method belongs here at all.
Use can use CollectionUtils.isEmpty()
but I suggest you to use just wellKnownConfig.getNetworkAuthenticationTargetIps().isEmpty() because you also have implemented returning an emptyset in the getter.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -136,18 +135,33 @@ private static Intent getAuthIntent(Map<String, String> parameters) {
         final Intent intent = new Intent();
         intent.setClass(getContext(), ConnectActivity.class);
         intent.setAction(ConnectUtils.LOGIN_ACTION);
+        String mccMnc = getMccMnc();
+        WellKnownAPI.WellKnownConfig wellKnownConfig = sdkProfile.getWellKnownConfig();
+        if (!TextUtils.isEmpty(mccMnc) && wellKnownConfig != null &&
+                !(isEmpty(wellKnownConfig.getNetworkAuthenticationTargetIps())
```
atanasdb: see comment below

connect/src/com/telenor/connect/ConnectSdk.java
gorandostanic: Done! (that was a good point :)

PR #62, opened 2017-07-04
Allow Operator Discovery to be performed via GSMA Operator Selection

gorandostanic:


jorunfa:
Is this pull request still relevant now with recent events (😢)?

PR #66, opened 2017-07-06
Fix the hotspot issue in the Android SDK

gorandostanic:


connect/src/com/telenor/connect/AbstractSdkProfile.java
```diff
@@ -166,6 +166,10 @@ protected void previewParameters(ParametersHolder parameters) {
                         && isEmpty(wellKnownConfig.getNetworkAuthenticationTargetUrls()))) {
                     parameters.add("login_hint", String.format("MCCMNC:%s", mccMnc));
         }
+
+        if (!ConnectSdk.isCellularDataNetworkConnected()) {
```
atanasdb: Look at shouldInterceptRequest
You could just do:
if (ConnectSdk.getCellularNetwork() == null && ConnectSdk.getWifiNetwork() != null) {
add no _seam
}

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -542,4 +542,28 @@ public static synchronized void initializeCommonComponents() {
         initalizeCellularNetwork();
         initalizeWiFiNetwork();
     }
+
+    public static boolean isCellularDataNetworkConnected() {
+        Validator.sdkInitialized();
+        ConnectivityManager connectivityManager =
```
atanasdb: we already have connectivityManager

connect/src/com/telenor/connect/AbstractSdkProfile.java
gorandostanic: I did that at first, hoping for the best, but it failed in my tests. Looks like ConnectSdk.getCellularNetwork() is not a reliable indicator for the availbaility of the cellular network.

For example, if you deactivate the cellular network (I switched the flight mode ON to simulate this), you won't be notified in the callback, meaning ConnectSdk.cellularNetwork will retain the old (stale) state despite the fact that the cellular network is unavailable nevertheless. You'll get your callback first when you reactivate the cellular network (e.g. by switching back the flight mode to OFF). In other words: once set to something != null, ConnectSdk.cellularNetwork will remain not null regardless of what you do.

You might need some changes in your force-HE patch to accommodate this.

atanasdb: No, it's fine. I mean this has been already considered in 

Check comments under 
https://confluence.skunk-works.no/confluence/display/Scale/Facilitating+Header+Enrichment+over+WiFi+in+MobileConnect+%28and+Connect+ID%29+SDK
and look for " We rely on the requestNetwork callbacks" for possible callbacks/adapters states

But I'm afraid we can't rely on that here, so maybe something like what you are doing will be necessary.

atanasdb: Even if you don't care whether Wifi is on/off I think it's nice to put the condition explicitely or at least explain it in a comment why you don't put it.

gorandostanic: To be honest, I'd prefer if you got rid of the Wi-Fi thing in the force-HE patch as well. I think you don't need it. Correct me if i am wrong, but I think the only thing you have to care about is the availability of the cellular network. If it is available, and if it is default, then you are lucky because you don't have to do anything. If it is not default, you force HE. In all other cases you do nothing. For example, I think there is an obvious bug in the current force-HE caused by taking into the equation a dependency which naturally isn't there. For example, if a phone is tethered via Bluetooth or otherwise, your wifi callback wouldn't fire => the app wouldn't behave as devised, would it?

atanasdb: I'm not sure I can call it a bug. Forced HE was planned to solve the WiFi + cellular use case as it's the most common and was never considered for Blouetooth/USB. Bluetooth and USB use cases are I guess not that common, but you are right that they could be solve similarly. Feel free to create JIRA issues for fixing that, or just fix it.  
Concerning this, let's leave it as it is now as you are right it will cover all cases. I only have the comment about ConnectivityManager.

connect/src/com/telenor/connect/ConnectSdk.java

connect/src/com/telenor/connect/AbstractSdkProfile.java
gorandostanic: The bug basically lies in the logical decomposition of the problem. The "opposite" of cellular connection is not WiFi, but a non-cellular connection. Besides, even with WiFi in picture, take a look at the following scenario: 
If you
1. start the application
2. turn Flight mode on
3. turn WiFi on
4. click on the CONNECT button
your patch will force HE over a disconnected cellular network. To tell the whole truth, you do catch IOException to mitigate this problem, but still, I think the code would be more efficient if you refrained from forcing anything through a network that you know is disconnected.

To illustrate my point, take a look at the patch where I tried to fix the problem: https://github.com/telenordigital/connect-android-sdk/pull/67 .

atanasdb: I'm not sure what we are discussing here :)
I agreed with you in my 2nd comment.
I just ask you to put in the description that with your patch you also add Bluetooth and USB functionality to Forced HE.
And also I don't agree it's a bug as this is by design and it's quite well documented in the link in my 2nd comment. The reason these comments are not part of the repo, but are only in Confluence is that because this repo is public.

atanasdb: But then again I think you might have problems with isCellularDataNetworkConnected.
If both wifi and cellular are on, this method will return false. That happens by design AFAIK, or at least happened with the device I tested with it. I recall now, this is why I took the callback approach, as it's the only one that worked for me.
If you can't fix isCellularDataNetworkConnected to behave sane, I'm afraid the hotspot could also not be solve in a similar way.

atanasdb: Goran, what do you think for something like : https://github.com/telenordigital/connect-android-sdk/commits/fix_hotspot_issue ?

gorandostanic: Are you sure about isCellularDataNetworkConnected? Works fine on my device, which indicates the issue you experience might not be by design anyway. But, since it's obviously you to have a device where this method does not work as it should, could you do me a favour? Could you test the same method on your device, but this time using PR 67? It contains another version of this method, which I tested and pushed it for review yesterday. If your test succeeds, I may replay the changes made in 67 in this PR.

atanasdb: Yes, if it works on your device, I'd say it's unpredictable then :/
Yes, I was testing with 67. You can see on Monday, with the white Android phone.
IIRC I saw some Stackflow posts describing that in the past, that this is an unreliable method for detecting connectivity.

atanasdb: Ok, I think the problem comes from 
networkInfo = connectivityManager.getNetworkInfo(ConnectivityManager.TYPE_MOBILE);
It does not work basically.
If the method is  
public static boolean isCellularDataNetworkConnected() {
  networkInfo = connectivityManager.getNetworkInfo(cellularNetwork);
  return (networkInfo != null) && networkInfo.isConnected();
}
it seems to work OK.

gorandostanic: Ok, so it seems it doesn't work on older devices :(

In that case, I suggest we resort to your smart solution for lower API levels, as a kind of legacy nobody will care about in the future, but keep the solution based on isCellularDataNetworkConnected if the API level is higher.

gorandostanic: For example, we may return true from it if the API level is low, to force your smart solution, but in all other cases we might return the real status of the cellular network.

PR #70, opened 2017-07-12
Extend Forced HE to support Bluetooth and USB

gorandostanic:
1. Avoid forcing HE on a disconnected (stale) cellular network
2. Make the patch WiFi-independent, making it possible to force HE over the tethered connection other that WiFi (e.g. Bluetooth and USB)

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -9,6 +9,7 @@ import android.content.pm.PackageManager;
 import android.net.ConnectivityManager;
 import android.net.Network;
 import android.net.NetworkCapabilities;
+import android.net.NetworkInfo;
 import android.net.NetworkRequest;
 import android.net.Uri;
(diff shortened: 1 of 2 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -518,17 +518,22 @@ public static synchronized void sdkReinitializeMobileConnect() {
      * Initialize components common to both Mobile Connect and ConnectID SDK profiles
      */
     private static synchronized void initializeCommonComponents() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
atanasdb: this has to be under getting connectivityManager
otherwise on older devices connectivityManager will be null and isCellularDataNetworkConnected is useless

gorandostanic: Yes, you are right. Done!

PR #74, opened 2017-07-18
Avoid SDK crash if the state changes since the last authentication

gorandostanic:


connect/src/com/telenor/connect/utils/ConnectUtils.java
```diff
@@ -25,12 +25,16 @@ public static void parseAuthCode(String callbackUrl, ConnectCallback callback) {
         Validator.notNullOrEmpty(callbackUrl, "callbackUrl");
 
         Uri uri = Uri.parse(callbackUrl);
-        Validator.validateAuthenticationState(uri.getQueryParameter("state"));
+        if (!Validator.validState(uri.getQueryParameter("state"))) {
```
atanasdb: are you sure u've included all the files in the commit

connect/src/com/telenor/connect/utils/ConnectUtils.java
```diff
@@ -49,4 +53,11 @@ public static void sendTokenStateChanged(boolean state) {
         intent.putExtra(LOGIN_STATE, state);
         localBroadcastManager.sendBroadcast(intent);
     }
+
+    private static Map<String, String> getErrorParams(String error, String errorDescription) {
```
atanasdb: I'm not sure this abstraction is worth it. It's one indirection more, and getErrorParams is used in one place only anyway.

connect/src/com/telenor/connect/utils/ConnectUtils.java
gorandostanic: What do you mean?

connect/src/com/telenor/connect/utils/ConnectUtils.java
gorandostanic: Done!

connect/src/com/telenor/connect/utils/ConnectUtils.java
atanasdb: Nothing ;).
I guess you will want to delete validateAuthenticationState now as it's not used any more.

connect/src/com/telenor/connect/utils/ConnectUtils.java
atanasdb: My github diff got a bit cluttered. You are actually calling getErrorParams twice. So maybe it make sense to leave it as it was the in the first commit. I'm sorry about that.

connect/src/com/telenor/connect/utils/ConnectUtils.java
gorandostanic: I'll do it in another patch (that I am currently working on)

connect/src/com/telenor/connect/utils/ConnectUtils.java
gorandostanic: I think I’ll leave it as it is, since it's in the spirit of this repo :)

PR #76, opened 2017-07-19
Scrap lastAuthState vairable

gorandostanic:
We don't need it, since it is already contained within the Uri the ConnectActivity
has been started with. Besides, not having this variable will remove the problems
we might have if developers contrive to call ConnectSdk.authorize() multiple times.
Some of the crash reports indicate that they are doing exactly that.


gorandostanic:
Now it is :)

PR #83, opened 2017-08-10
Change naming of payment link

stoutbeard:
After creating a new transaction, the webserver was passing a link to
the payment ui via a json field called `location`. Apparently, this has
caused 3rd party developers to misunderstand, and redirect their users
to the `Location` header returned from payment. This leads them to be
presented with the wrong payment UI.


jorunfa:
Looks fine, I guess. As a note: Do you know if the Payment part of the SDK is generally in use by third party devs?

stoutbeard:
I think they mostly use it as an example implementation

PR #85, opened 2017-08-28
Add logOut feature

jorunfa:


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -336,7 +336,13 @@ public static synchronized boolean isInitialized() {
...
     public static void logout() {
         Validator.sdkInitialized();
-        sdkProfile.getConnectIdService().revokeTokens(getContext());
+
+        if (sdkProfile instanceof MobileConnectSdkProfile) {
+            sdkProfile.getConnectIdService().revokeTokens(getContext());
+            return;
+        }
+
+        sdkProfile.getConnectIdService().logOut(getContext());
```
atanasdb: Is it an option to add logout as part of SdkProfile interface.
Then you can just call sdkProfile.logout here,
for mc logout you could just: connectIdService.revokeTokens()
for connectid: connectIdService.logout

jorunfa: Hmm... maybe... but I'm not sure that would make for cleaner code since the revoke and logOut functions needs the access token which, at least for connect id, is managed by `ConnectIdService` class (and this class is currently calling `connectApi.revokeToken(...)`.

atanasdb: but that's fine , connectApi will point to a MobileConnectAPIAdapter in case of an MC profile and to ConnectAPI in case of a ConnectId. Or I'm not getting your concern correctly?

jorunfa:
@atanasdb: Added your commit to the PR.

connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -170,6 +175,42 @@ public void failure(RetrofitError error) {
...
+        if (accessToken == null) {
+            Log.w(ConnectUtils.LOG_TAG, "Trying to log out when user is already logged out.");
+            return;
+        }
+
+        Date accessTokenExpirationTime = getAccessTokenExpirationTime();
+        Date now = new Date();
+
+        boolean accessTokenHasExpired = now.after(accessTokenExpirationTime);
+        if (accessTokenHasExpired) {
```
ulfcarlin: This check does not take into consider that the access token may have been revoked in other ways, or that the refresh token may still be valid. I think a safer approach would be always to try to refresh the access token first. If the revoke call is successful, it is very likely that the new access token will be valid immediately after. If revoke fails, the client does not have access anymore, so it can be considered logged out.

connect/src/com/telenor/connect/id/ConnectIdService.java
```diff
@@ -170,6 +175,42 @@ public void failure(RetrofitError error) {
...
+                public void failure(RetrofitError error) {
+                    Log.e(ConnectUtils.LOG_TAG, "Failed to call logout with access token on API. accessToken=" + accessToken , error);
+                }
+            });
+        }
+
+        String refreshToken = getRefreshToken();
+        if (!TextUtils.isEmpty(refreshToken)) {
+            revokeRefreshToken(refreshToken);
+        }
```
ulfcarlin: I think the access tokens should be revoked as well

jorunfa:
@ulfcarlin: More like this?

Issue #91, opened 2018-01-29
ConnectLoginButton launches Chrome custom tab in new Task


rasmusohrstig:
The Chrome custom tab launched by ConnectLoginButton always launches in a new Android Task. 

This means that our login UI will from the user's point of view seem be a separate "app" in the app switcher, which can be confusing.

We will also have to set the launchMode="singleTask" of the Activity that launches the Custom tab in order to prevent the callback to start another instance of that Activity in the same task as the Custom tab.

It would be nice if ConnectLoginButton had a way to chose between new task and same task. This could for example be done with a constructor, or a setter, or in some other way (builder pattern etc).

PR #93, opened 2018-01-30
Adding a setter for the custom tab in same/new task behavior

rasmusohrstig:
A setter lets you launch the custom tabs in the current Task instead of in a new Task. The custom tabs still launch in a new Task by default. See issue #91.


jorunfa:
Thanks for all the great contributions 🎉😎

jorunfa:
Looking at this part again for #137, I'm starting to think the default should always be false, and in indeed that the initial implementation was erroneous (too quick copy paste from https://developer.chrome.com/multidevice/android/customtabs).

Unless I hear some objections I plan to remove the toggle, and never use `Intent.FLAG_ACTIVITY_NEW_TASK`; so that we always open the custom tab in the same task.

PR #107, opened 2018-04-04
Send analytics data

atanasdb:


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -657,6 +730,23 @@ public void onAvailable(Network network) {
...
+        GoogleApiAvailability googleAPI = GoogleApiAvailability.getInstance();
+        if (googleAPI.isGooglePlayServicesAvailable(getContext()) == ConnectionResult.SUCCESS) {
+            new Thread(new Runnable() {
+                public void run() {
+                    AdvertisingIdClient.Info adInfo = null;
+                    try {
+                        adInfo = AdvertisingIdClient.getAdvertisingIdInfo(getContext());
+                        advertisingId = adInfo.getId();
+                    } catch (Exception e) {
+                        Log.e(ConnectUtils.LOG_TAG, "Failed to read advertisment id", e);
```
jorunfa: If it is possible that this will fail due to blocked permissions/settings (I am not sure here), then maybe we should just log with a warning instead of error?

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -666,4 +756,30 @@ public static Network getCellularNetwork() {
...
+    private static String getApplicationName() {
+        try {
+            return getContext().getApplicationInfo().loadLabel(getContext().getPackageManager()).toString();
+        } catch (Exception e) {
+            Log.e(ConnectUtils.LOG_TAG, "Failed to read application name", e);
+            return "";
+        }
+    }
+
+    private static String getApplcationVersion() {
```
jorunfa: `getApplcationVersion`->
`getApplicationVersion`

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -103,6 +117,11 @@ public static SdkProfile getSdkProfile() {
         return sdkProfile;
     }
 
+    public static void beforeAuthentication() {
+        logSessionId = UUID.randomUUID().toString();
+        tsLoginButtonClicked = System.currentTimeMillis();
+    }
+
     public static synchronized void authenticate(
```
jorunfa: For this `authenticate` method, should we do `parameters.put("lsi", getLogSessionId())`, so it gets passed along?

connect-id-example/proguard-rules.pro
```diff
@@ -40,4 +40,21 @@
...
+ }
+
+-keepnames @com.google.android.gms.common.annotation.KeepName class *
+-keepclassmembernames class * {
+    @com.google.android.gms.common.annotation.KeepName *;
+}
+
+-keepnames class * implements android.os.Parcelable {
+    public static final ** CREATOR;
+}
```
jorunfa: N00b question: How do you get this list of things for the `proguard-rules.pro` file?

atanasdb: Not an easy task :)
Some googling, trial and error. Usually library devs publish their proguard-rules or repos like
https://github.com/krschultz/android-proguard-snippets/tree/master/libraries

connect/src/com/telenor/connect/ConnectSdk.java
atanasdb: Done. Put in the shared authenticate method

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -125,6 +126,7 @@ private Uri getAuthorizeUri() {
         }
 
         parameters.put("state", connectStore.generateSessionStateParam());
+        parameters.put("lsi", ConnectSdk.getLogSessionId());
```
jorunfa: ConnectLoginButton calls `authenticate`, so I guess we can remove this now. I guess we could/should move the same `"state"` logic (above call) to the authenticate method as well, but your choice if you want to do that now or not.

atanasdb: nope, we cab't becauseChrome Custom Tabs does not call authenticate....
not super happy with the existing design, but you can put it in the tech debt list

PR #113, opened 2018-04-17
Fix ConnectLoginButton context leak

simonnorberg:
LeakCanary reports a leak every time we use the ConnectLoginButton in our app:

> In se.telenor.mytelenor.alphatest:1.5.6.135.1:135.
> * com.telenor.app.mytelenor.ui.NewLoginActivity has leaked:
> * GC ROOT android.support.customtabs.CustomTabsClient$2.this$0 (anonymous subclass of android.support.customtabs.ICustomTabsCallback$Stub)
> * references android.support.customtabs.CustomTabsServiceConnection$1.this$0 (anonymous subclass of android.support.customtabs.CustomTabsClient)
> * references com.telenor.connect.ui.ConnectLoginButton$1.this$0 (anonymous subclass of android.support.customtabs.CustomTabsServiceConnection)
> * references com.telenor.connect.ui.ConnectLoginButton.mContext
> * leaks com.telenor.app.mytelenor.ui.NewLoginActivity instance

It looks like the anonymous subclass of CustomTabsServiceConnection leaks a reference to the Activity context. Changing to a private static class with a weak reference to the button fixes the problem for us. I've not been able to reproduce the issue with your example app but it would be great if you would consider this change anyway.

### DIFF
--- a/connect/src/com/telenor/connect/ui/ConnectLoginButton.java
+++ b/connect/src/com/telenor/connect/ui/ConnectLoginButton.java
@@ -60,7 +60,7 @@ public class ConnectLoginButton extends ConnectWebViewLoginButton {
         super(context, attributeSet);
         connectStore = new ConnectStore(getContext());
         setText(R.string.com_telenor_connect_login_button_text);
-        connection = new MyCustomTabsServiceConnection(new WeakReference<>(this));
+        connection = new WeakReferenceCustomTabsServiceConnection(new WeakReference<>(this));
         onClickListener = new LoginClickListener();
         setOnClickListener(onClickListener);
     }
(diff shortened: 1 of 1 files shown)


jorunfa:
Thanks for PR, will have a look when I have the chance :) 💃

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -210,4 +186,57 @@ public void onClick(View v) {
         }
 
     }
+
+    private static class MyCustomTabsServiceConnection extends CustomTabsServiceConnection {
```
jorunfa: Rename
`MyCustomTabsServiceConnection` to
`WeakReferenceCustomTabsServiceConnection`?

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -210,4 +186,57 @@ public void onClick(View v) {
...
+                session.mayLaunchUrl(PRE_FETCH_URL, null, null);
+            }
+        }
+
+        @Override
+        public void onServiceDisconnected(ComponentName name) {
+        }
+    }
+
+    private static class MyCustomTabsCallback extends CustomTabsCallback {
```
jorunfa: Consider renaming similarly to above comment.

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
simonnorberg: Sure, I have renamed them.

jorunfa:
LGTM. Will merge it in. Going to make a new release after #115 is closed.

PR #114, opened 2018-04-18
Update gradle and change deprecated compile to implementation

jorunfa:
References: https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html?utm_source=android-studio#new_configurations
https://medium.com/mindorks/implementation-vs-api-in-gradle-3-0-494c817a6fa

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,12 +2,11 @@
 
 buildscript {
     repositories {
-        mavenLocal()
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.1'
         classpath 'com.novoda:bintray-release:0.7.0'
         classpath 'com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.8.2'
 
--- a/connect/build.gradle
+++ b/connect/build.gradle
@@ -5,7 +5,7 @@ project.group = 'com.telenor.connect'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion '26.0.2'
+    buildToolsVersion '27.0.3'
     defaultConfig {
         minSdkVersion 15
         targetSdkVersion 27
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -2,12 +2,11 @@
 
 buildscript {
     repositories {
-        mavenLocal()
         google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.1'
```
andereld: One of the links in your commit message mentions Gradle 3.4; is there a particular reason to move to 3.1 rather than 3.4?

connect/build.gradle
```diff
@@ -34,14 +34,14 @@ android {
...
-    compile 'com.android.support:support-v4:27.1.0'
-    compile 'com.google.android.gms:play-services-basement:8.3.0'
+    implementation 'com.android.support:appcompat-v7:27.1.1'
+    implementation 'com.squareup.retrofit:retrofit:1.9.0'
+    implementation 'com.squareup.okhttp3:okhttp:3.8.1'
+    implementation 'com.nimbusds:nimbus-jose-jwt:3.10'
+    implementation 'com.android.support:customtabs:27.1.1'
+    implementation 'com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0'
+    implementation 'com.android.support:support-v4:27.1.1'
+    implementation 'com.google.android.gms:play-services-basement:8.3.0'
```
andereld: From what I've read, it seems that the compile -> implementation change was a breaking change in Gradle 3.0. Has the build configuration been broken for some time?

build.gradle
jorunfa: Good question. This is just the plugin, which should be the latest, while the actual Gradle being used is 4.4.
https://github.com/telenordigital/connect-android-sdk/pull/114/files#diff-fd9d61e37d939c941e7a62077f38d62dR6
https://developer.android.com/studio/releases/gradle-plugin.html#updating-plugin

connect/build.gradle
jorunfa: The warning I've been getting was that it was deprecated and support would be removed by the end of 2018.

PR #115, opened 2018-04-20
Improve web error feedback

jorunfa:
This change will give users who are failing to reach our servers with WebView browsers a clearer error message and code, and a details view. The idea is to preserve the UX while also giving information which can be given to us for debugging (e.g. ina screenshot).

new_device_1_overview
new_device_2_details


connect/res/values/strings.xml
```diff
@@ -7,4 +7,6 @@
     <string name="com_telenor_connect_network_settings"><u>Network settings</u></string>
     <string name="com_telenor_mobile_connect_login_button_text">Sign in with Mobile Connect</string>
     <string name="com_telenor_authorization_cancelled" translatable="false">Authorisation cancelled (initialization failed)!</string>
+    <string name="show_details">Show details</string>
```
andereld: Will these be translated and added in a later commit?

connect/src/com/telenor/connect/ui/ConnectWebViewClient.java
```diff
@@ -185,9 +186,12 @@ public WebResourceResponse fetchUrlTroughCellular(String originalUrl) {
         return null;
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.M)
```
andereld: Err, why is Marshmallow required for this functionality? What happens to all of the users on older systems?

connect/res/values/strings.xml
andereld: Okey-dokey.

connect/src/com/telenor/connect/ui/ConnectWebViewClient.java
andereld: I see, thanks.

andereld:
I understand the motivation; the issue is that the text doesn't look scrollable to me. I noticed now that it also looks a bit weird with the text so far to the left when everything else is centred; maybe reducing the width a bit would help.

Anyway, seeing as this text is only shown when the user clicks ‘Show details’, I think it would be reasonable to show the entire text, pushing the ‘Try again’ button etc. down on the page. So then the entire page will scroll rather than just the URL, or you can click ‘Hide details’ to bring things back to how they were. What do you think?

jorunfa:
Sounds reasonable. Not sure what the best UX would be, but your example sounds very reasonable, and has the added benefit of giving more debug info to us.

jorunfa:
@andereld: The view now behaves like you suggested it should (if I didn't misunderstand).

Issue #119, opened 2018-05-09
Logging out is unreliable


rasmusohrstig:
If we call ConnectSdk.logout() we expect to get a confirmation that we have logged out. This happens, but the problem is that it happens even when the network request to the revoke endpoint fails. 

We can recreate this error in our app by enabling airplane mode before clicking the logout button. It is probably possible to recreate it in the same way in the example project, but I have not tried that yet.

Based on the code it looks like it would be fairly straightforward to fix the bug by removing the call to clearTokensAndNotify() from ConnectIdService.revokeTokens() and adding appropriate implementations of success(Response response) and failure(RetrofitError error) in the ResponseCallback used in ConnectIdService.revokeToken()

The current implementation of failure(RetrofitError error) also logs the still valid token to logcat, if I'm reading the code correctly, which seems like a minor security flaw.

rasmusohrstig:
So just to clarify: the root cause of the bug is that the SDK ignores the result of the network request to "/oauth/revoke".

jorunfa:
See your point, and if I can remember correctly, we had a (security) discussion before agreeing on this implementation. I think the gist of it was that oauth/openid connect allows for best effort logging out/revoking of tokens, and that the user experience would suffer if, for an example, the user could not log out when you have a spotty mobile connection or on a plane. We were not absolutely sure it was the better call when we made it.

Is there a specific reason why you need a callback, which could help in reevaluating?

The logging is not intentional, will have a look at that.

rasmusohrstig:
Well, this came up during testing. The refresh token stored in Chrome will still be valid, which means you can just re-open the app, click the login button, get logged in automatically and continue using the app.

I think this is a security issue for a user who is using an app on a shared device that supports Chrome custom tabs.

I don't think it's possible to tell Chrome or the Chrome custom tab to delete the cookies, so there does not seem to be a workaround that will fully wipe the still valid tokens from the device.

From a developer's point of view I think it would be useful to get more fine-grained callbacks, at least when using Chrome custom tabs.

jorunfa:
See your point - I think we can add that. Not sure how soon we can prioritise this one, but it shouldn't be very hard, as you mentioned.

jorunfa:
I started looking at this, but I think I would like to challenge you on how you would like this to work - what kind of interface would you like and what expectations do you have.

There are a few network calls which all can fail, in `ConnectIdService.java`:
[...]

PR #124, opened 2018-06-08
Set useStaging in sdkInitialize

simonnorberg:
The `useStaging` boolean is read from the manifest and cannot be changed during runtime. This is a problem for us because we have to build two versions of our app when we do internal testing. Here is a suggestion for moving the staging configuration to the `sdkInitialize()` method instead, it would save us both time and confusion.

### DIFF
--- a/connect-id-example/src/main/java/com/telenor/connect/connectidexample/ExampleApplication.java
+++ b/connect-id-example/src/main/java/com/telenor/connect/connectidexample/ExampleApplication.java
@@ -8,6 +8,6 @@ public class ExampleApplication extends Application {
     @Override
     public void onCreate() {
         super.onCreate();
-        ConnectSdk.sdkInitialize(getApplicationContext(), true);
+        ConnectSdk.sdkInitialize(getApplicationContext());
     }
 }
--- a/connect-id-example/src/main/java/com/telenor/connect/connectidexample/ExampleApplication.java
+++ b/connect-id-example/src/main/java/com/telenor/connect/connectidexample/ExampleApplication.java
@@ -8,6 +8,6 @@ public class ExampleApplication extends Application {
     @Override
     public void onCreate() {
         super.onCreate();
-        ConnectSdk.sdkInitialize(getApplicationContext());
+        ConnectSdk.sdkInitialize(getApplicationContext(), true);
     }
 }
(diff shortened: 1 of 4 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -326,12 +321,18 @@ public static void logout() {
     }
 
     public static synchronized void sdkInitialize(Context applicationContext) {
+        sdkInitialize(applicationContext, false);
```
jorunfa: Change `false` -> `true`.

connect-id-example/src/main/java/com/telenor/connect/connectidexample/ExampleApplication.java
```diff
@@ -8,6 +8,6 @@
     @Override
     public void onCreate() {
         super.onCreate();
-        ConnectSdk.sdkInitialize(getApplicationContext());
+        ConnectSdk.sdkInitialize(getApplicationContext(), true);
```
jorunfa: Remove `, true`.

connect/src/com/telenor/connect/ConnectSdk.java
simonnorberg: Ok, but that breaks the test case `getConnectApiUrlReturnsProductionByDefault()`. Should I change the test as well?

jorunfa: You can delete that whole test. The method used to be in the `ConnectSdk` class, but since it got moved to `ConnectUrlHelper` it doesn't really belong there anymore. And the method is pretty trivial, so should be fine to delete it.

simonnorberg: Done.

jorunfa:
Thanks for the contribution! 🎉 
For now I'm planning to make a change to the docs for this change, and resolve #119, and then make a new release.

simonnorberg:
Awesome 👍

jorunfa:
Docs pull request: https://github.com/telenordigital/id-docs.telenordigital.com/pull/9

jorunfa:
Released: https://github.com/telenordigital/connect-android-sdk/releases/tag/v1.7.0

PR #128, opened 2018-06-13
Update the id token

rasmusohrstig:
Always update the locally stored id token when a new one is received.

The ConnectStore.update() method becomes identical to the ConnectStore.set() method when we always update the id token, so we only need the set() method.

### DIFF
--- a/connect/src/com/telenor/connect/id/ConnectIdService.java
+++ b/connect/src/com/telenor/connect/id/ConnectIdService.java
@@ -239,7 +239,7 @@ public class ConnectIdService {
                                 = HeadersDateUtil.extractDate(response.getHeaders());
                         ConnectTokens connectTokens
                                 = new ConnectTokens(connectTokensTO, serverTimestamp);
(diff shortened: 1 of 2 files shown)


jorunfa:
If I'm not mistaken you would then loose the ID token after refreshing the access token once (say after 1 hour), as you don't get new ones on refresh.

rasmusohrstig:
Yeah this will not work. The api gives us a null id token when we refresh our access token.

It looks like the ID token can not be updated using the api. In that case there is no way to fix this issue in the SDK code.

ulfcarlin:
The authorization server is by design only returning an ID token at authentication time and not at refresh token time. This functionality is a policy decision made for the Telenor Digital authorization which is in line with the OpenID Connect specification, even though the specification also allows ID tokens to be issued at refresh token time (see http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse). Therefore, this pull request does not make sense, and should therefore not be considered a candidate for being merged.

Issue #127, opened 2018-06-12
Upgrade to Retrofit 2


simonnorberg:
I really recommend upgrading to Retrofit 2, it is great! 😃 Any plans for that?

I can submit a PR if you want.

jorunfa:
I've had it in the back of my head that it would probably be a good idea, but there hasn't been any requests for it, so I haven't been able to prioritise it. A PR would be very welcome if you feel like it :D

PR #129, opened 2018-06-21
Update to Retrofit 2

simonnorberg:
Hi, here's a suggestion for updating to Retrofit 2. See  #127
We're using Retrofit 2 in our app and it would be great if the Connect SDK uses the same.

The biggest change is that the Retrofit interfaces returns `Call<T>` objects and we now pass in callbacks to `call.enqueue()`. We also get error status codes in `onResponse` and have to check `response.isSuccessful()` there.

I've tried to keep the diff minimal, so I may not have formatted the code perfectly.

Example ProGuard rules are updated according to documentation for each library.

The WellKnownAPI mock methods in TestHelper are removed because they were unused and I don't know how to make them work with Retrofit 2.

I tested this with our app and it seems to work fine, but obviously it needs testing from you.

### DIFF
--- a/connect/src/com/telenor/connect/WellKnownAPI.java
+++ b/connect/src/com/telenor/connect/WellKnownAPI.java
@@ -9,17 +9,17 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
-import retrofit.Callback;
-import retrofit.http.GET;
-import retrofit.http.Headers;
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Headers;
 
 public interface WellKnownAPI {
 
-    public static final String OPENID_CONFIGURATION_PATH = "/.well-known/openid-configuration";
+    String OPENID_CONFIGURATION_PATH = ".well-known/openid-configuration/";
 
     @Headers("Content-Type: application/json")
     @GET("/")
-    void getWellKnownConfig(Callback<WellKnownConfig> callback);
+    Call<WellKnownConfig> getWellKnownConfig();
 
     class WellKnownConfig implements Parcelable {
 
--- a/connect/src/com/telenor/connect/WellKnownAPI.java
+++ b/connect/src/com/telenor/connect/WellKnownAPI.java
@@ -98,6 +98,14 @@ public interface WellKnownAPI {
         @SerializedName("telenordigital_sdk_analytics_endpoint")
         private String analyticsEndpoint;
         public String getAnalyticsEndpoint() {
+            if (analyticsEndpoint != null) {
+                try {
+                    if ('/' != analyticsEndpoint.charAt(analyticsEndpoint.length() - 1)) {
+                        return analyticsEndpoint + "/";
+                    }
+                } catch (IndexOutOfBoundsException ignored) {
+                }
+            }
             return analyticsEndpoint;
         }
     }
--- a/connect/src/com/telenor/connect/WellKnownAPI.java
+++ b/connect/src/com/telenor/connect/WellKnownAPI.java
@@ -15,10 +15,8 @@ import retrofit2.http.Headers;
 
 public interface WellKnownAPI {
 
-    String OPENID_CONFIGURATION_PATH = ".well-known/openid-configuration/";
-
     @Headers("Content-Type: application/json")
-    @GET("/")
+    @GET("/oauth/.well-known/openid-configuration")
     Call<WellKnownConfig> getWellKnownConfig();
 
     class WellKnownConfig implements Parcelable {
--- a/connect-id-example/proguard-rules.pro
+++ b/connect-id-example/proguard-rules.pro
@@ -1,43 +1,35 @@
 # Connect SDK
 -keep class com.telenor.** { *; }
 
-# Retrofit 1.X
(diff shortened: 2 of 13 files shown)


jorunfa:
Awesome pull request! Looks promising. I'll have to give Retrofit 2 docs a look over to get up to speed and do some testing. Will try to get it prioritised.

connect/src/com/telenor/connect/WellKnownAPI.java
```diff
@@ -9,17 +9,17 @@
...
-import retrofit.http.GET;
-import retrofit.http.Headers;
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Headers;
 
 public interface WellKnownAPI {
 
-    public static final String OPENID_CONFIGURATION_PATH = "/.well-known/openid-configuration";
+    String OPENID_CONFIGURATION_PATH = ".well-known/openid-configuration/";
```
jorunfa: Need to cut the trailing slash for the endpoint:
`".well-known/openid-configuration/"` ->
`".well-known/openid-configuration"`

simonnorberg: Ah yes, I see the problem.

The path in the baseUrl `"oauth/.well-known/openid-configuration/"` is overridden by the path in the annotation `@GET("/")`

But the baseUrl must end with `"/"` or it will crash when you set it in `RestHelper.buildApi()`
https://github.com/simonnorberg/connect-android-sdk/blob/retrofit2/connect/src/com/telenor/connect/utils/RestHelper.java#L79

See documentation
https://square.github.io/retrofit/2.x/retrofit/retrofit2/Retrofit.Builder.html#baseUrl-okhttp3.HttpUrl-

I changed to have the path in the annotation in WellknownAPI instead. Similar to in ConnectAPI.
`@GET("/oauth/.well-known/openid-configuration")`

jorunfa:
I got this probably related exception:
```java
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: com.telenor.connect.connectidexample, PID: 28597
                  java.lang.IllegalArgumentException: baseUrl must end in /: https://connect.staging.telenordigital.com/id/extapi/analytics
                      at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:515)
                      at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:458)
                      at com.telenor.connect.utils.RestHelper.buildApi(RestHelper.java:79)
                      at com.telenor.connect.utils.RestHelper.getApi(RestHelper.java:45)
                      at com.telenor.connect.utils.RestHelper.getAnalyticsApi(RestHelper.java:39)
                      at com.telenor.connect.ConnectSdk.sendAnalyticsData(ConnectSdk.java:234)
                      at com.telenor.connect.ConnectSdk.access$100(ConnectSdk.java:53)
                      at com.telenor.connect.ConnectSdk$1.onSuccess(ConnectSdk.java:212)
                      at com.telenor.connect.id.ConnectIdService$1.onResponse(ConnectIdService.java:95)
                      at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
                      at android.os.Handler.handleCallback(Handler.java:789)
                      at android.os.Handler.dispatchMessage(Handler.java:98)
                      at android.os.Looper.loop(Looper.java:164)
                      at android.app.ActivityThread.main(ActivityThread.java:6938)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
```

Could you apply a similar kind of ointment to this url while you're at it? 🙏

simonnorberg:
Yes, that's the exception I got. This endpoint is a little tricky as you get it from a JSON response.

Would this be an acceptable solution?

https://github.com/simonnorberg/connect-android-sdk/blob/master/connect/src/com/telenor/connect/WellKnownAPI.java#L102-L104

``` java
public String getAnalyticsEndpoint() {
    if (analyticsEndpoint != null) {
        try {
            if ('/' != analyticsEndpoint.charAt(analyticsEndpoint.length() - 1)) {
                return analyticsEndpoint + "/";
            }
        } catch (IndexOutOfBoundsException ignored) {
        }
    }
    return analyticsEndpoint;
}
```

jorunfa:
I don't think it's a good idea to add trailing `/` to url's that are not supposed to have that. But I can accept the solution and try to fix the root problem.

simonnorberg:
> I don't think it's a good idea to add trailing / to url's that are not supposed to have that.

I agree!

> But I can accept the solution and try to fix the root problem.

Ok 👍

jorunfa:
Sorry this has taken ages, but I finally got the time to come to a conclusion. The proper solution would probably be to add the trailing `/` to the url in `openid-configuration`, but that would affect all older versions of the SDK. So for now at least I think we'll go with your graceful hack. [Another possible way would be using `@Url`](https://stackoverflow.com/a/32559579), but that would be hacky too. There are worse buried in the project ⚰️. Thanks again! 🎉

Will try to make a new release next week.

simonnorberg:
Great, thanks!

PR #131, opened 2018-09-20
Improve updateTokens and getValidAccessToken callback

### LINKED ISSUE #130: Problem with getValidAccessToken()
We call `ConnectSdk.getValidAccessToken(AccessTokenCallback callback)` in `onResume()` in our launch activity to check if we are logged in with ConnectID. If we get a `onSuccess()` callback we assume that we are logged in and have a valid IdToken. However, this is not always the case. Sometimes `ConnectSdk.getIdToken().getSerializedSignedJwt()` and `ConnectSdk.getIdToken().getSubject()` returns null if we call them in the AccessTokenCallback onSuccess.

Is this working as intended? Our understanding is that if we have a valid access token we should always have an IdToken? Is there a recommended way to verify that a user is logged in?

We are using version 1.7.1.

jorunfa:
Fixes #130.

Calling these methods no longer risks an unchecked exception.

The new callback interface looks like this:

```
public interface AccessTokenCallback {
    void success(String accessToken);
    void unsuccessfulResult(Response response, boolean userDataRemoved);
    void failure(Call<ConnectTokensTO> call, Throwable error);
    void noSignedInUser();
}
```

Upon failure, which is common for network failure, one can retry the method called.

Jira: IDLIVE-352


connect-id-example/src/main/java/com/telenor/connect/connectidexample/SignedInActivity.java
```diff
@@ -48,4 +44,39 @@ private void goToLogin() {
...
+            @Override
+            public void success(String accessToken) {
+                Toast.makeText(SignedInActivity.this,
+                        "accessToken can now be used to access user sensitive resources",
+                        Toast.LENGTH_SHORT)
+                        .show();
+            }
+
+            @Override
+            public void unsuccessfulResult(Response response, boolean userWasCleared) {
```
rasmusohrstig: The parameter "userWasCleared" could be misinterpreted as meaning something like "user was granted security clearance". Could perhaps be renamed to something like "userDataRemoved".

jorunfa: Sounds good 👍📝

connect-id-example/src/main/java/com/telenor/connect/connectidexample/SignedInActivity.java
```diff
@@ -48,4 +44,39 @@ private void goToLogin() {
...
+            public void success(String accessToken) {
+                Toast.makeText(SignedInActivity.this,
+                        "accessToken can now be used to access user sensitive resources",
+                        Toast.LENGTH_SHORT)
+                        .show();
+            }
+
+            @Override
+            public void unsuccessfulResult(Response response, boolean userDataRemoved) {
+                // a 4xx response will sign out any signed in user
```
ulfcarlin: Since the response often contains useful information for debugging, consider including some code which accesses the response.

PR #132, opened 2018-10-03
Delete .gitignore

895aungye:
Recently we seem to be falling back on the SDK:s built in web solution (ConnectActivity, WebView) on most devices almost all the time. We have seen this problem on Samsung Galaxy S7, S8 with Android 8 and Google Pixel 1 with Android 9.

I have tried to warm up the Chrome custom tab service and waiting before attaching the button to the layout, but it does not seem to help.

The if clause on line 80 in CustomTabsHelper seems to always be false. So in other words the PackageManager is not able to resolve the Intent.


jorunfa:
Why do you want to delete the .gitignore file?

895aungye:
Sorry sir

Issue #133, opened 2018-10-03
Chrome custom tab no longer works (most of the time)


rasmusohrstig:
Recently we seem to be falling back on the SDK:s built in web solution (ConnectActivity, WebView) on most devices almost all the time. We have seen this problem on Samsung Galaxy S7, S8 with Android 8 and Google Pixel 1 with Android 9.

I have tried to warm up the Chrome custom tab service and waiting before attaching the button to the layout, but it does not seem to help.

The if clause on line 80 in CustomTabsHelper seems to always be false. So in other words the PackageManager is not able to resolve the Intent.

rasmusohrstig:
By they way, it would be a good idea to move the Intent resolution from View.onAttachedToWindow() to when the user presses the button. That would make it possible to warm up Chrome custom tabs before the Intent resolution code runs.

simonnorberg:
We're using custom tabs from support library 28 with the new androidx package.
https://developer.android.com/jetpack/androidx/

`implementation "androidx.browser:browser:1.0.0"`

I think Gradle forces Connect SDK to use this version as well.

jorunfa:
There is one known problem I've seen before:
![image](https://user-images.githubusercontent.com/1413275/46468380-8ac39a80-c7d0-11e8-8cd2-43ab773a9235.png)

The resolved activity for the intent is the ResolverActivity which, if I'm not mistaken, is the activity for choosing which app you want to open the intent in. This happens if you have multiple apps capable of opening the same uri. Do you have multiple apps using the same uri scheme installed on the device?

Can you debug this line and check? https://github.com/telenordigital/connect-android-sdk/blob/master/connect/src/com/telenor/connect/ui/ConnectLoginButton.java#L78

jorunfa:
I see we're also using support library 27.something in the SDK. Can bump it to 28.

jorunfa:
Can you try
```
$ git clone git@github.com:telenordigital/connect-android-sdk.git
$ cd connect-android-sdk/
$ git checkout custom-tabs
$ open -a /Applications/Android\ Studio.app .
(run id-example)
```
And see if that works? And if doing `$ git checkout master` or `git checkout v1.8.0` and running the example app isn't working?

rasmusohrstig:
I tested your branch and it actually works. I also tried migrating to AndroidX in the same branch and it still works after that. So that's good.

[...]

PR #134, opened 2018-10-03
Delete ExampleApplication.java

895aungye:


895aungye:
Recently we seem to be falling back on the SDK:s built in web solution (ConnectActivity, WebView) on most devices almost all the time. We have seen this problem on Samsung Galaxy S7, S8 with Android 8 and Google Pixel 1 with Android 9.

I have tried to warm up the Chrome custom tab service and waiting before attaching the button to the layout, but it does not seem to help.

The if clause on line 80 in CustomTabsHelper seems to always be false. So in other words the PackageManager is not able to resolve the Intent.

jorunfa:
Thank you for your input. Is this pull request intended to be an issue (is this a duplicate of #133)? I will try to resolve that issue as soon as possible.
(I assume deleting `Delete ExampleApplication.java` doesn't resolve it?)

PR #136, opened 2018-11-02
Add Instant verification

jorunfa:
1. This change is huge. Sorry about that.
2. There is lots of non-obvious moving of existing code. (`ConnectLoginButton` -> `ConnectLoginButtonWithoutProgressBar`. The `ConnectLoginButton` is new.)
3. Lots of existing code have lacklustre testing. A fix for this has not been prioritised.
4. Quite a bit of manual testing have been done.
5. Known missing: Flag or way to disable the feature.


connect/res/layout/com_telenor_connect_login_button_with_progress_bar.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    >
+    <com.telenor.connect.ui.ConnectLoginButtonWithoutProgressBar
```
froeseth: The file says *with* progress bar, while this line of code says *without*. It's not immediately apparent to me why they're the opposite.

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -359,19 +428,13 @@ public void onFailure(Call<WellKnownAPI.WellKnownConfig> call, Throwable error)
...
+        HeLogic.initializeNetworks(context, useStaging);
         initializeAdvertisingId(context);
         isInitialized = true;
-        tsSdkInitiliazation = System.currentTimeMillis();
+        tsSdkInitialization = System.currentTimeMillis();
     }
 
-
-    public static String getMccMnc() {
+    private static String getMccMnc() {
```
froeseth: Two abbreviations in one method, well done ;D 
Not sure if it's neccessary, but consider spelling out one or two of the abbreviations to hint at people like me what this is about.

connect/src/com/telenor/connect/headerenrichment/GetHeaderEnrichmentGifTask.java
```diff
@@ -0,0 +1,75 @@
...
+            return null;
+        }
+
+        HeTokenResponse heTokenResponse = getHeTokenResponse(getTokenResponse);
+        if (heTokenResponse == null) {
+            return null;
+        }
+
+        String getGifResponse = MobileDataFetcher.fetchUrlTroughCellular(heTokenResponse.getGifUrl());
+        return getGifResponse != null ? heTokenResponse : null;
```
froeseth: Reverse conditional and use `== null`? Negations are harder to read (in this case, `not null` harder to read than `null`).

connect/src/com/telenor/connect/headerenrichment/GetHeaderEnrichmentGifTask.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private static HeTokenResponse getHeTokenResponse(String getTokenResponse) {
```
froeseth: Hmmm, not sure about this parameter name. It's used in a get method, but the "get" in the name is from a get request...?

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+
+import com.telenor.connect.ConnectSdk;
+import com.telenor.connect.utils.ConnectUrlHelper;
+
+import java.util.Date;
+import java.util.Map;
+
+public class HeLogic {
+    private static final long HE_TOKEN_TIMEOUT_MILLISECONDS = 10_000;
+    private static final boolean cantDirectNetworkTraffic = Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP;
```
froeseth: Reverse this to be positive? That is, let parameter indicate that we _can_ direct network traffic.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+import android.os.Build;
+
+import com.telenor.connect.ConnectSdk;
+import com.telenor.connect.utils.ConnectUrlHelper;
+
+import java.util.Date;
+import java.util.Map;
+
+public class HeLogic {
+    private static final long HE_TOKEN_TIMEOUT_MILLISECONDS = 10_000;
```
froeseth: As an outsider, I'm curious as to why we use 10 000 here. However, not sure if it makes sense to add a comment. What do you think? It may make it easier to contribute to the project.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+    private static HeTokenCallback heTokenCallback;
+    private static boolean isHeTokenRequestOngoing;
+    private static HeTokenResponse heTokenResponse;
+    private static ConnectivityManager connectivityManager;
+    private static volatile Network cellularNetwork;
+    private static volatile Network defaultNetwork;
+
+    public static void initializeNetworks(Context context, boolean useStaging) {
+        connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        if (connectivityManager != null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
froeseth: Consider naming the conditional like you've done in other classes. I like the context it gives.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+                        public void onAvailable(Network network) {
+                            cellularNetwork = network;
+                            boolean noSignedInUser = ConnectSdk.getAccessToken() == null;
+                            if (noSignedInUser) {
+                                initializeHeaderEnrichment(useStaging, ConnectSdk.getLogSessionId());
+                            }
+                        }
+                    }
+            );
+        } catch (SecurityException e) {
```
froeseth: No need for logging here?

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+            cellularNetwork = null;
+        }
+    }
+
+    private static void initializeHeaderEnrichment(boolean useStaging, String logSessionId) {
```
froeseth: Perhaps not necessary to spell out HE here since it's not done elsewhere(?), and also, this is the HE logic class. 

I know, I know! I'm nitpicking...

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    public static void handleHeToken(
+            final Map<String, String> parameters,
+            final ShowLoadingCallback showLoadingCallback,
+            final HeTokenCallback heTokenCallback,
+            final String logSessionId,
+            final boolean useStaging) {
+        boolean finishedUnSuccessfully = !heTokenSuccess && !isHeTokenRequestOngoing;
```
froeseth: Also here is a negative parameter. Consider renaming to `finishedSuccessfully`.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -0,0 +1,210 @@
...
+        NetworkInfo networkInfo;
+        if (cantDirectNetworkTraffic) {
+            networkInfo = connectivityManager.getNetworkInfo(ConnectivityManager.TYPE_MOBILE);
+        } else {
+            if (cellularNetwork == null) {
+                return false;
+            }
+            networkInfo = connectivityManager.getNetworkInfo(cellularNetwork);
+        }
+        return (networkInfo != null) && networkInfo.isConnected();
```
froeseth: You have parentheses around the first conditional here, and this differs from elsewhere (see f ex line 198). Is there a reason for the parentheses here?

connect/src/com/telenor/connect/headerenrichment/HeTokenResponse.java
```diff
@@ -0,0 +1,27 @@
+package com.telenor.connect.headerenrichment;
+
+import java.util.Date;
+
+public class HeTokenResponse {
```
froeseth: Have you considered using Lombok for some of the boiler plate code? E.g. for both constructor and getters and setters in this class.

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map;
+
+import static java.net.HttpURLConnection.HTTP_MOVED_PERM;
+import static java.net.HttpURLConnection.HTTP_MOVED_TEMP;
+import static java.net.HttpURLConnection.HTTP_SEE_OTHER;
+
+public class MobileDataFetcher {
+
+    private static final int MAX_REDIRECTS_TO_FOLLOW_FOR_HE = 5;
+
```
froeseth: Remove one of the two empty lines?

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static java.net.HttpURLConnection.HTTP_MOVED_PERM;
+import static java.net.HttpURLConnection.HTTP_MOVED_TEMP;
+import static java.net.HttpURLConnection.HTTP_SEE_OTHER;
+
+public class MobileDataFetcher {
+
+    private static final int MAX_REDIRECTS_TO_FOLLOW_FOR_HE = 5;
```
froeseth: I'm curious, why do we need a max for redirects?

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+
+
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    public static String fetchUrlTroughCellular(String url) {
```
froeseth: Missing "h" in `Through`.

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+        } while (attempts <= MAX_REDIRECTS_TO_FOLLOW_FOR_HE);
+        return null;
+    }
+
+    private static Map<String, String> getHeadersAsCommaJoinedList(Map<String, List<String>> headerFields) {
+        Map<String, String> result = new HashMap<>();
+        for (String header : headerFields.keySet()) {
+            List<String> strings = headerFields.get(header);
+            if (strings == null || strings.isEmpty()) {
```
froeseth: Not sure how much string processing is done and if it makes sense, but there is a method com.google.common.base.Strings.isNullOrEmpty(). You've most likely seen it in the connect repository.

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+            return false;
+        }
+        if (!wellKnownConfig.getNetworkAuthenticationTargetUrls().isEmpty()) {
+            for (String urlPrefix : wellKnownConfig.getNetworkAuthenticationTargetUrls()) {
+                if (url.contains(urlPrefix)) {
+                    return true;
+                }
+            }
+            return false;
+        } else {
```
froeseth: Unneccesary else as all above returns anyways.

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -1,215 +1,148 @@
...
-                    .getConnectApiUrl()
-                    .newBuilder()
-                    .addPathSegment("id")
-                    .addPathSegment("android-sdk-prefetch-static-resources")
-                    .build()
-                    .uri()
-                    .toString()
-    );
+    private void init() {
+        inflate(getContext(), R.layout.com_telenor_connect_login_button_with_progress_bar, this);
```
froeseth: Here's the contrast again that I asked about in another class also. The class contains a button `ConnectLoginButtonWithoutProgressBar`, but here I see a parameter containing ...`_with_progress_bar`.

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -1,215 +1,148 @@
...
-        onClickListener = new LoginClickListener();
-        setOnClickListener(onClickListener);
+    private void setLoading(boolean loading) {
+        progressBar.setVisibility(loading ? VISIBLE : INVISIBLE);
+        loginButton.setEnabled(!loading);
     }
 
-    public OnClickListener getOnClickListener() {
-        return onClickListener;
+    public ConnectLoginButton(Context context, AttributeSet attrs) {
```
froeseth: Isn't the convention to gather constructors at the top of the class?

connect/src/com/telenor/connect/ui/ConnectLoginButtonWithoutProgressBar.java
```diff
@@ -0,0 +1,169 @@
...
+
+import com.telenor.connect.BrowserType;
+import com.telenor.connect.ConnectSdk;
+import com.telenor.connect.R;
+import com.telenor.connect.utils.ConnectUrlHelper;
+import com.telenor.connect.utils.CustomTabsHelper;
+
+import java.lang.ref.WeakReference;
+
+public class ConnectLoginButtonWithoutProgressBar extends ConnectWebViewLoginButton {
```
froeseth: Somewhat confusing that this class extends a class named _web view_ while at the same time containing fields named _custom tabs_. I assume it's common functionality? Can the common functionality be named something else? (Disclaimer: I don't know a whole lot about veb views vs custom tabs, so perhaps it makes sense to mix the two :P )

froeseth: So... On authentication, the button falls back on web view if custom tabs not supported? Perhaps documentation of the class could be helpful in explaining this? Doesn't have to be more than one sentence.

connect/src/com/telenor/connect/headerenrichment/GetHeaderEnrichmentGifTask.java
jorunfa: Changed to guard statement, and also above, to be inline with our styleguide: https://kb.skunk-works.no/display/eng/Java+Coding+Guidelines#JavaCodingGuidelines-Guardstatements

connect/src/com/telenor/connect/headerenrichment/GetHeaderEnrichmentGifTask.java
jorunfa: I agree, yeah overlapping get meaning. Does it look better now? (2248787)

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Done: aff02c9.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Sorry, it's very non-obvious, but this is moved code from https://github.com/telenordigital/connect-android-sdk/blob/master/connect/src/com/telenor/connect/ConnectSdk.java#L573, and not new code.

I found the commit where it was added: 99257872c0a6d0d23f6320171af87a0ee5bdb948. So apparently it's a known bug for 6.0 and something that we wouldn't otherwise expect...
I guess we could add a comment in the code or something.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Same argument as before: never reverted.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Good question - I don't know :p (not the original author). Looking at the original commit I don't see any reason for it: 04be260611697b0c3f96d1797b40778aaa2102e1

connect/src/com/telenor/connect/headerenrichment/HeTokenResponse.java
jorunfa: Not done a full assessment, no. I'd be willing to consider it, but would in that case like to the change in a separate pull request.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: d322499

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Yes. 3b1a6ad

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Not sure; moved code. The original commit included a link to the original design doc: https://kb.skunk-works.no/display/Scale/Facilitating+Header+Enrichment+over+WiFi+in+MobileConnect+%28and+Connect+ID%29+SDK

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Lol, yes. Fixed: 0a24591

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    public static String fetchUrlTroughCellular(String url) {
+        WebResourceResponse webResourceResponse = MobileDataFetcher.fetchWebResourceResponse(url, false);
+        if (webResourceResponse == null) {
+            return null;
+        }
+
+        int statusCode = webResourceResponse.getStatusCode();
+        if (statusCode < 200 || statusCode >= 300) {
```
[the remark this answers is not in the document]
jorunfa: Fixed: e75c339

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Could use `TextUtils.isEmpty(string)`, Android's built-in version. I think my argument is that I find it faster and safer to write out the null check, since it's short, and since there are hundred different string null check util methods one can use. If I did use that, someone not familiar with that method would have to look it up to verify it does a null check, although I guess you could reasonably assume it does. What do you think?

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: You are correct - points to Gryffindor ;). (btw; it was moved code) 65b0b8a

connect/src/com/telenor/connect/ui/AuthenticationButton.java
```diff
@@ -0,0 +1,36 @@
...
+
+    void setAcrValues(String... acrValues);
+
+    void setAcrValues(ArrayList<String> acrValues);
+
+    void setLoginScopeTokens(String... scopeTokens);
+
+    void setLoginScopeTokens(ArrayList<String> scopeTokens);
+
+    void addLoginParameters(Map<String, String> parameters);
```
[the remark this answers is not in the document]
jorunfa: Renamed method so that it better reflects what it does. 8193f22

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
jorunfa: Ref comment https://github.com/telenordigital/connect-android-sdk/pull/136#discussion_r240228262, is this more clear now?

connect/src/com/telenor/connect/headerenrichment/HeTokenResponse.java
froeseth: Good idea to keep it a separate commit. If at all important enough to change...

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
froeseth: To me, Strings.isNullOrEmpty() is very descriptive. That night be because I'm used to it. However, considering it checks for two things in one method (null or empty), it's more readable than two conditionals. 
I have no strong opinions, two conditionals work just fine :)

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
froeseth: Yes, it's more clear. 
However, is it clear enough for other new readers of the code without explanation or thorough reading of the code? Consider if it makes sense to add a sentence as comment in the class. No biggie, drop it if it doesn't make sense.

connect/src/com/telenor/connect/ConnectSdk.java
froeseth: I guess many developers don't know all abbreviations from the telecom world, so then it makes sense to spell it out. I don't think this is an abbreviation used much in the Connect sphere, is it? In that case, perhaps spell it out. Unless it makes sense for us to learn that abbreviation, then keep it :)

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
froeseth: Don't know... I guess it's implied that the number should be "long enough to succeed and short enough for users to wait it out if it should fail/time out". My best suggestion would be to reuse what you said: "Timeout is a best guess at what would be long enough to succeed and short enough for users to wait it out if it should fail/time out, and do a normal flow."

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
froeseth: I see. Perhaps then use "canNotDirect..." to make the negation more obvious (for readability).

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
froeseth: Ok. (As opposed to the other comment, the naming of this parameter can stay the same. The negation has its own word and camel hump, namely "Un", which makes the negation part more readable.)

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Or maybe I'm mixing some things up here... This appears to be checking that a list is not null and not empty, not a string, as I thought, when I wrote the first comment.

So neither Strings.isNullOrEmpty(String) nor TextUtils.isEmpty(String) would work in this case.

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
jorunfa: Is `ConnectCustomTabLoginButton` a better name than `ConnectLoginButtonWithoutProgressBar` perhaps?

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
froeseth: I see that now, good point :O Perhaps the parameter should be renamed to reflect that, e.g. `stringList` or `headerList`.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Changed: 4f43e29

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
jorunfa: Fixed: b27acf7

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
froeseth: Yes, that makes sense to me and sounds like a good idea. I like that it relates to the class that it extends, namely `ConnectWebViewLoginButton`. Also, it avoids (or at least hides :) ) the confusion with the progress bar.

connect/src/com/telenor/connect/headerenrichment/MobileDataFetcher.java
jorunfa: Changed: 1e09acf

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
jorunfa: Fixed: 2772159

jorunfa:
@froeseth, fyi I added 36ef984 and 06e0f86 before merging.

PR #137, opened 2018-11-14
Otp autofill

jorunfa:
Using _SmsRetriever API_.

For both WebView and custom tabs.


connect/src/com/telenor/connect/sms/AppSignatureHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public AppSignatureHelper(Context context) {
+        super(context);
+    }
+
+    /**
+     * Get all the app signatures for the current package
+     * @return
+     */
+    public ArrayList<String> getAppSignatures() {
```
vlarsen: I usually like return types and variable types of collection classes to be the interface rather than the concrete implementation; unless you rely specifically on the implementation where the value is returned to.

connect/src/com/telenor/connect/sms/SmsPinParseUtil.java
```diff
@@ -7,6 +7,8 @@
 
 public class SmsPinParseUtil {
 
+    private static final Pattern PIN_PATTERN = Pattern.compile("<#>.*([0-9]{4}).*.{11}");
```
vlarsen: Is this pattern the entire message, so should have anchoring with ^ and $?

vlarsen: The "4" here is the length of the pin? So if we change this we must change the pattern?

connect/src/com/telenor/connect/sms/AppSignatureHelper.java
jorunfa: Agree, but this is actual borrowed Google code: https://github.com/googlesamples/android-credentials/blob/master/sms-verification/android/app/src/main/java/com/google/samples/smartlock/sms_verify/AppSignatureHelper.java

I think perhaps I'm leaning towards just deleting this file, and just referring to the Google documentation, in our docs (https://developers.google.com/identity/sms-retriever/verify#computing_your_apps_hash_string). The code should never be in any production product anyways.

connect/src/com/telenor/connect/sms/SmsPinParseUtil.java
jorunfa: Kinda/perhaps. I know some BUs prefix their sms with the cost of the sms. It shouldn't be picked up by Android then, as it doesn't match the requirements, but I didn't want to explicitly block it, in case it did work. Don't know if there is any point to it.

jorunfa: Yes, yes, and yes.

vlarsen: I'm a bit worried that matching the hash with just plain '.' might be too permissive. As it is a base64 fragment, the following should work
```
[A-Za-z0-9\+\/]{11}
````
(if I quoted correctly...)

PR #150, opened 2019-02-07
Enable instant verification in all environments

jorunfa:


connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -63,10 +63,6 @@ public void onAvailable(Network network) {
         }
     }
 
-    private static boolean instantVerificationEnabled() {
-        return ConnectSdk.useStaging();
-    }
-
     private static void initializeHe(boolean useStaging, String logSessionId) {
```
paulrene: He -> HeaderEnrichment

jorunfa: Like so? d65baeb

paulrene:
lgtm

PR #152, opened 2019-02-12
Generate new logSessionId in more scenarios

jorunfa:
Will now generate a new logSessionId on aborts, old sessions, and all
success scenarios (even when we fail to send analytics).

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -47,6 +47,7 @@ import org.json.JSONObject;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -117,10 +119,20 @@ public void done() {
...
 
-    private static void setButtonClickedTimestamp() {
+    private static void handleButtonClickedAnalytics() {
+        updateLogSessionIdIfTooOld();
         tsLoginButtonClicked = System.currentTimeMillis();
     }
 
+    private static void updateLogSessionIdIfTooOld() {
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.MINUTE, -10);
```
froeseth: Is it usefult to add information here about why we use 10 minutes? (Was this the one that was discussed earlier, where the minimum of other settings was 10 minutes?)

jorunfa: I'll add some context :) 👍

jorunfa: Better now?

Issue #145, opened 2019-01-17
Memory leak in webview/ConnectActivity when otp autofill doesn't get used


jorunfa:
```
E/ActivityThread: Activity com.telenor.connect.ui.ConnectActivity has leaked IntentReceiver com.telenor.connect.sms.SmsBroadcastReceiver@e8c5823 that was originally registered here. Are you missing a call to unregisterReceiver()?
    android.app.IntentReceiverLeaked: Activity com.telenor.connect.ui.ConnectActivity has leaked IntentReceiver com.telenor.connect.sms.SmsBroadcastReceiver@e8c5823 that was originally registered here. Are you missing a call to unregisterReceiver()?
        at android.app.LoadedApk$ReceiverDispatcher.<init>(LoadedApk.java:1355)
        at android.app.LoadedApk.getReceiverDispatcher(LoadedApk.java:1120)
        at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1428)
        at android.app.ContextImpl.registerReceiver(ContextImpl.java:1401)
        at android.app.ContextImpl.registerReceiver(ContextImpl.java:1389)
        at android.content.ContextWrapper.registerReceiver(ContextWrapper.java:622)
        at com.telenor.connect.ui.ConnectWebViewClient.<init>(ConnectWebViewClient.java:67)
        at com.telenor.connect.ui.ConnectWebFragment.onCreateView(ConnectWebFragment.java:64)
        at android.support.v4.app.Fragment.performCreateView(Fragment.java:2439)
        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1460)
        at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1784)
        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1852)
        at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:802)
        at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2625)
        at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2411)
        at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2366)
        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2273)
[...]

PR #156, opened 2019-03-05
Fix SmsBroadcastReceiver leaking activity

simonnorberg:
We get an activity leak when we login with ChromeCustomTab in SDK 1.9.5. Possibly related to #145 

> com.telenor.app.mytelenor.ui.LoginActivity has leaked:
> thread Thread.!(<Java Local>)! (named 'Answers Events Handler1')
> ↳ static ExecutorUtils$1$1.!($class$dexCache)!
> ↳ DexCache.!(runtimeInternalObjects)!
> ↳ array Object[].!([1376])!
> ↳ static ConnectSdk.!(smsBroadcastReceiver)!
> ↳ SmsBroadcastReceiver.!(smsHandler)!
> ↳ ConnectSdk$2.!(val$activity)! (anonymous implementation of com.telenor.connect.sms.SmsHandler)
> ↳ LoginActivity

It looks like smsBroadcastReceiver is only unregistered if an SMS is received?
https://github.com/telenordigital/connect-android-sdk/blob/master/connect/src/com/telenor/connect/ConnectSdk.java#L152-L165

I'm not sure it is the best solution but unregistering/clearing smsBroadcastReceiver after successful login seems to fix the leak.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -156,7 +156,7 @@ public final class ConnectSdk {
                 if (pin == null) {
                     return;
                 }
-                getContext().unregisterReceiver(smsBroadcastReceiver);
+                safeUnregisterAndRemoveBroadcastReceiver();
                 String url = ConnectUrlHelper.getSubmitPinUrl(pin);
                 Uri uri = Uri.parse(url);
                 launchUrlInCustomTab(activity, session, uri);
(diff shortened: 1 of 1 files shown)


jorunfa:
I think you are completely right that I/we missed that scenario. I like the way you're doing it here, but I think you have to surround the thing in a try/catch-ignore here, in case the user did get an SMS and it was unregistered already

```
try {
    unregisterReceiver(broadcastReceiver);
} catch (IllegalArgumentException ignore) {}
```

And if you wrap the change in a method called `safeUnregisterAndRemoveBroadcastReceiver` I would like to use that method for the other unregister as well (in the BroadcastReceiver).

Might it be that the revert did nothing about the memory leak, then?
I suggest I merge your change when it is ready; I revert the revert; and then I can make an alpha release which you could test?

simonnorberg:
I added the method `safeUnregisterAndRemoveBroadcastReceiver` and it works for me but I'm unable to test the `receivedSms` callback.

The weak reference in Custom tabs login button is still needed. That was a different leak trace.

jorunfa:
Will try to get this released before the weekend.

simonnorberg:
Thanks!

jorunfa:
https://github.com/telenordigital/connect-android-sdk/releases/tag/v1.9.6

jorunfa:
There is also an updated instant verification release (if you didn't already test the other one): https://github.com/telenordigital/connect-android-sdk/releases/tag/v1.11.2-beta

simonnorberg:
Awesome!

jorunfa:
@simonnorberg: Did you get a chance to test the beta, by any chance?
If you can log in once without any hiccups then it's probably fine.

If you don't have the time, then that is something I can relate to 😏.

simonnorberg:
@jorunfa 
I've tried 1.11.3-beta now and I can login without any problems.

jorunfa:
Thanks for checking out!

PR #163, opened 2019-03-28
Add analytics data about user usage of mobile data dialog

SerhiiBovtriuk:


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -338,6 +346,20 @@ public static void sendAnalyticsData(Throwable e) {
             Log.e(ConnectUtils.LOG_TAG, "Exception making exception json", e1);
         }
 
+        JSONObject mobileDialogData = new JSONObject();
+        try {
+            mobileDialogData
```
jorunfa: I don't like that we're creating this json object in almost the same way twice (i.e. we're repeating ourselves, and not DRY). Either create a helper method, or something else clever.

Although it's sent as JSON, maybe we should make a POJO class, and then that will be serialised to json when sent. EnableMobileDataDialogAnalytics.java, perhaps. I'm not sure.

connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -20,6 +20,8 @@
 import com.telenor.connect.headerenrichment.HeLogic;
 import com.telenor.connect.headerenrichment.ShowLoadingCallback;
 import com.telenor.connect.id.Claims;
+import com.telenor.connect.utils.EnableMobileDataDialogAnalytics;
+
```
jorunfa: Seems like a lot of empty new lines between these imports. Trim?

jorunfa:
Merge at any time, captain 👨‍✈️

PR #164, opened 2019-04-02
Rename EnableMobileData -> TurnOnMobileData

jorunfa:
To make the difference between what the feature does and whether or not
it is enabled more clear.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -40,7 +40,7 @@ import com.telenor.connect.ui.ConnectWebFragment;
 import com.telenor.connect.ui.ConnectWebViewLoginButton;
 import com.telenor.connect.utils.ConnectUrlHelper;
 import com.telenor.connect.utils.ConnectUtils;
-import com.telenor.connect.utils.TurnOnMobileDataAnalytics;
+import com.telenor.connect.utils.TurnOnMobileDataDialogAnalytics;
 import com.telenor.connect.utils.RestHelper;
 import com.telenor.connect.utils.Validator;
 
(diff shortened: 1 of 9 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -411,7 +412,7 @@ public static String getClientId() {
 
     public static boolean showMobileDataDialog() {
```
SerhiiBovtriuk: I guess method also needs to be renamed.

connect/src/com/telenor/connect/utils/TurnOnMobileDataAnalytics.java
```diff
@@ -7,14 +7,14 @@
 
 import java.io.Serializable;
 
-public class EnableMobileDataDialogAnalytics implements Serializable {
+public class TurnOnMobileDataAnalytics implements Serializable {
```
SerhiiBovtriuk: Maybe it should be "TurnOnMobileData**Dialog**Analytics"?

connect/src/com/telenor/connect/ConnectSdk.java
jorunfa: Fixed in e3fa8e2.

jorunfa: Good catch.

connect/src/com/telenor/connect/utils/TurnOnMobileDataAnalytics.java
jorunfa: It was intentional, but I think perhaps I've changed my mind and agree with you. The dialog is core to the feature.

PR #170, opened 2019-04-26
Add support for multiple brands

SerhiiBovtriuk:


jorunfa:
Is the button expected to stay the same, or is it supposed to have a different look for different brands?
Could either add all of that (if that's the plan), or rename the pull request (add support for multiple brand endpoints)

SerhiiBovtriuk:
I didn't get clear message from anybody about the button status for now, so this question is still in the air. I will update this when there will be more details.

connect/src/com/telenor/connect/id/IdProvider.java
```diff
@@ -0,0 +1,18 @@
+package com.telenor.connect.id;
+
+public enum IdProvider {
+    CONNECT_ID("connect.telenordigital.com", "connect.staging.telenordigital.com"),
+    TELENOR_ID("example.com", "staging-telenorid.com"); // TODO: production is not set yet
```
vlarsen: I think the decision is to not use toplevel domain for the IDP, but rather signin.staging-telenorid.com

SylviaBrudeli: Yes, you are right Vidar!

Issue #168, opened 2019-04-23
NullPointerException: Attempt to invoke interface method 'java.util.Set java.util.Map.entrySet()'


XarBpeak:
**Fatal Exception: java.lang.NullPointerException**

```
: Attempt to invoke interface method 'java.util.Set java.util.Map.entrySet()' on a null object reference
       at com.telenor.connect.ui.ConnectActivity.onError(ConnectActivity.java)
       at com.telenor.connect.ConnectSdk$1.onError(ConnectSdk.java)
       at com.telenor.connect.id.ConnectIdService$1.onResponse(ConnectIdService.java)
       at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java)
       at android.os.Handler.handleCallback(Handler.java:815)
```

PR #171, opened 2019-04-29
Fix possible NPE when the received object is null

SerhiiBovtriuk:
Details: https://github.com/telenordigital/connect-android-sdk/issues/168


connect/src/com/telenor/connect/ui/ConnectActivity.java
```diff
@@ -70,10 +70,12 @@ public void onConfigurationChanged(Configuration newConfig) {
     @Override
     public void onSuccess(Object successData) {
         if (ConnectSdk.isConfidentialClient()) {
-            Map<String, String> authCodeData = (Map<String, String>) successData;
             Intent intent = new Intent();
-            for (Map.Entry<String, String> entry : authCodeData.entrySet()) {
-                intent.putExtra(entry.getKey(), entry.getValue());
+            if (successData instanceof Map) {
```
jorunfa: Can't have a success where the result doesn't contain the information needed.

PR #176, opened 2019-05-31
Investigate ActivityNotFoundException while working with custom tabs

SerhiiBovtriuk:


SerhiiBovtriuk:
Not sure here, but null activity is a valid case anyway, so better to lock it out.

Issue #166, opened 2019-04-23
ConnectLoginButton InflateException


XarBpeak:
> android.view.InflateException
: Binary XML file line #604: Error inflating class com.telenor.connect.ui.ConnectLoginButton

where at line 604 
```
<com.telenor.connect.ui.ConnectLoginButton
                        android:id="@+id/btn_connectLogin"
                        style="?android:attr/buttonStyleSmall"
                        android:layout_width="0dp"
                        android:layout_height="wrap_content"
                        android:layout_marginLeft="10dp"
                        android:layout_marginStart="10dp"
                        android:layout_weight="5"
                        android:background="@drawable/selectable_button"
                        android:paddingLeft="5dp"
                        android:paddingRight="5dp"
                        android:visibility="gone"
                        android:text="@string/proceed"
                        android:textColor="@color/white"
                        android:textStyle="bold" />

```

**Caused by java.lang.reflect.InvocationTargetException**
      ```
 at java.lang.reflect.Constructor.constructNative(Constructor.java)
       at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
       at android.view.LayoutInflater.createView(LayoutInflater.java:594)
       at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:696)
```

**NullPointerException**
       ```
at com.telenor.connect.ui.**ConnectLoginButton.onVisibilityChanged**(ConnectLoginButton.java)
       at android.view.View.dispatchVisibilityChanged(View.java:7959)
       at android.view.ViewGroup.dispatchVisibilityChanged(ViewGroup.java:1060)
```

SerhiiBovtriuk:
Hello, @XarBpeak. 

It's not completely clear what are you trying to do here, can you provide more information?

XarBpeak:
This issue is only **coming in 5.1.1**   inflating ConnectLoginButton crashes.
<com.telenor.connect.ui.ConnectLoginButton

SerhiiBovtriuk:
@XarBpeak what exactly is version 5.1.1? 

Can you, please, provide the whole error stack trace and code where the error is happening?

XarBpeak:
5.1.1 means lollipop . Simply try on any device whose version is 5.1.1 it will occur . right now I don't have that device

SerhiiBovtriuk:
@XarBpeak thanks for the response.

This should be fixed in next SDK release.

SerhiiBovtriuk:
[...]

PR #177, opened 2019-06-13
Fix SDK button crashes if it's in "invisible" or "gone" state in Android 5

SerhiiBovtriuk:
Related to https://github.com/telenordigital/connect-android-sdk/issues/166


connect/src/com/telenor/connect/ui/ConnectLoginButton.java
```diff
@@ -122,6 +122,9 @@ public ConnectLoginButton(Context context, AttributeSet attrs, int defStyleAttr,
     @Override
     protected void onVisibilityChanged(@NonNull View changedView, int visibility) {
         super.onVisibilityChanged(changedView, visibility);
+        if (loginButton == null || loginButton.getActivity() == null) {
+            return;
```
metatelenor: Should we log this as an error?

SerhiiBovtriuk: No, because it's not an error. It's more about specific Android version behaviour.

PR #178, opened 2019-06-21
Fix npe in well known config

SerhiiBovtriuk:
The issue can happen, when this part is called from a different thread,
however, it's hard to reproduce. The solution is to make a workaround
to prevent crashes, while investigating deeply.

### DIFF
--- a/connect/src/com/telenor/connect/ConnectSdk.java
+++ b/connect/src/com/telenor/connect/ConnectSdk.java
@@ -67,7 +67,7 @@ public final class ConnectSdk {
     private static ConnectIdService connectIdService;
     private static Context context;
     private static boolean confidentialClient;
-    private static WellKnownAPI.WellKnownConfig wellKnownConfig;
+    private static volatile WellKnownAPI.WellKnownConfig wellKnownConfig;
     private static volatile boolean isInitialized = false;
     private static String clientId;
     private static String redirectUri;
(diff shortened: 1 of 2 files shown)


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -67,7 +67,7 @@
     private static ConnectIdService connectIdService;
     private static Context context;
     private static boolean confidentialClient;
-    private static WellKnownAPI.WellKnownConfig wellKnownConfig;
+    private static volatile WellKnownAPI.WellKnownConfig wellKnownConfig;
```
vlarsen: is this variable now unused?

connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -329,16 +329,15 @@ private static void sendAnalyticsData() {
...
-    public static void sendAnalyticsData(Throwable e) {
-        WellKnownAPI.WellKnownConfig wellKnownConfig = getWellKnownConfig();
-        if (wellKnownConfig == null) {
-            return;
-        }
-        String analyticsEndpoint = wellKnownConfig.getAnalyticsEndpoint();
-        if (analyticsEndpoint == null) {
+    public static void sendAnalyticsData(Throwable throwable) {
+        String analyticsEndpoint;
+        // This is bad practise. The source of the issue must be found.
```
metatelenor: Typo. Please change practise to practice :)

connect/src/com/telenor/connect/ConnectSdk.java
SerhiiBovtriuk: @vlarsen Variable is still used in other places.

connect/src/com/telenor/connect/ConnectSdk.java
SerhiiBovtriuk: @metatelenor Valid point, will fix that.

jorunfa:
@SerhiiBovtriuk, @metatelenor, @vlarsen: Sorry slow reply. The `analyticsEndpoint` can be null if the request has not finished or if the request failed if it's the first time ever fetching it. I.e.:
https://github.com/telenordigital/connect-android-sdk/blob/master/connect/src/com/telenor/connect/ConnectSdk.java#L517
https://github.com/telenordigital/connect-android-sdk/blob/master/connect/src/com/telenor/connect/ConnectSdk.java#L523

The UI does not stop the user from signing in or doing whatever, even if this request is still loading or has failed. A bug/feature of sorts - I think the idea was not to block the user even if this failed or delay the user from signing in. The old forced header enrichment (before instant verification) also depends on this endpoint, so it's also a source of that sometimes not working as expected (instant verification does not depend on this request to get any config).

PR #182, opened 2019-08-07
Add new style for Telenor ID button

SerhiiBovtriuk:


connect/src/com/telenor/connect/ui/ConnectWebViewLoginButton.java
```diff
@@ -35,8 +39,12 @@
 
     public ConnectWebViewLoginButton(Context context, AttributeSet attributeSet) {
         super(context, attributeSet);
-        String brandName = ConnectSdk.getIdProvider().getName();
-        setText(getResources().getString(R.string.com_telenor_connect_login_button_text, brandName));
+        IdProvider brand = ConnectSdk.getIdProvider();
+        if (brand == IdProvider.TELENOR_ID) {
+            AssetManager am = context.getApplicationContext().getAssets();
+            setTypeface(Typeface.createFromAsset(am, "fonts/telenorregularwebfont.ttf"));
```
vlarsen: which typeface will be set for other IdProvider brands?

SerhiiBovtriuk: @vlarsen By default we have "sans-serif-light". This, probably, should be unified for all our interfaces in future. 

However, I can set Telenor font as default now, the only issue with that is it might look not perfect on older screens.

vlarsen: I guess that is a designer decision; and should be addressed in a future change

PR #191, opened 2019-11-07
[android sdk] Add manage {brand name} link

SerhiiBovtriuk:
Add the option to call a method in sdk to open
"Manage {brand name}" page properly inside of
Chrome Custom Tabs or default browser instead.


connect-id-example/src/main/java/com/telenor/connect/connectidexample/SignInActivity.java
```diff
@@ -15,13 +17,21 @@
 public class SignInActivity extends AppCompatActivity {
 
     ConnectLoginButton loginButton;
+    Button randomButton;
```
vlarsen: why is this called randomButton? Does it do random stuff?

PR #192, opened 2019-11-25
Remove option to use fallback components directly

SerhiiBovtriuk:
We expect SDK users to have a single entry point to the SDK
and we provide additional fallback options in case if that entry
point fails. Before it was possible to use fallback options directly,
that was allowing the usage of some bad practises. Now this option
will be disabled.

**This is a breaking change.**

### DIFF
--- a/connect/src/com/telenor/connect/ui/ConnectWebViewLoginButton.java
+++ b/connect/src/com/telenor/connect/ui/ConnectWebViewLoginButton.java
@@ -23,16 +23,14 @@ import java.util.Map;
 
 import androidx.annotation.NonNull;
 
-public class ConnectWebViewLoginButton extends ConnectButton implements AuthenticationButton {
-
-    public static final int NO_CUSTOM_LAYOUT = -1;
+class ConnectWebViewLoginButton extends ConnectButton implements AuthenticationButton {
 
     private ArrayList<String> acrValues;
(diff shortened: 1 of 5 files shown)


connect/src/com/telenor/connect/ui/ConnectWebViewLoginButton.java
```diff
@@ -23,16 +23,14 @@
...
-    public static final int NO_CUSTOM_LAYOUT = -1;
+class ConnectWebViewLoginButton extends ConnectButton implements AuthenticationButton {
 
     private ArrayList<String> acrValues;
     private Map<String, String> loginParameters;
     private ArrayList<String> loginScopeTokens;
     private int requestCode = 0xa987;
     private Claims claims;
-    private int customLoadingLayout = NO_CUSTOM_LAYOUT;
+    private int customLoadingLayout = -1;
```
vlarsen: When removing the constant, can you remove this variable as well? In any case I'm a bit worried that you now have -1 two different places in the code.

SerhiiBovtriuk: @vlarsen there is a big question if this part is needed at all. 

However, the reason for removing constant is to remove dependency that is now not accessible. Removing options for custom layout is a task of different scope that can be done later.

PR #194, opened 2019-12-06
Add temporary fix for HE traffic

SerhiiBovtriuk:
Now SDK is doing HE request each time when there is a network switch. That results in massive ---`prepare` and `inject` traffic towards Header Enrichment. 

This is a temporary fix for the issue, that will set the flag `heWasInitialized ` to true whenever HE prepare-inject was initialised.

Proper fix is scheduled to be done in future. It should include rewrite of network handling logic and solving threading issue while reading results from HE.

### DIFF
--- a/connect/src/com/telenor/connect/headerenrichment/HeLogic.java
+++ b/connect/src/com/telenor/connect/headerenrichment/HeLogic.java
@@ -24,6 +24,7 @@ public class HeLogic {
     private static final long HE_TOKEN_TIMEOUT_MILLISECONDS = 10_000; // Timeout is a best guess
     // at what would be long enough to succeed and short enough for users to wait it out if
     // it should fail/time out, and do a normal flow
+    private static boolean heWasInitialized = false;
     private static boolean heTokenSuccess = true;
     private static HeTokenCallback heTokenCallback;
     private static boolean isHeTokenRequestOngoing;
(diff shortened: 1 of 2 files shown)


vlarsen:
How is this a breaking change?

SerhiiBovtriuk:
> How is this a breaking change?

Missclick, marked as "bug".

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -196,6 +197,9 @@ private static void handleHeTokenResult(HeTokenResponse heTokenResponse) {
         if (heTokenCallback != null) {
             heTokenCallback.done();
         }
+        if (!failedToGetToken() && !heWasExecutedAndReceivedSuccessResponse) {
```
vlarsen: Don't really need to check if this was set to true previously, before setting it to true again. Might even just write `heWasExecBlahBlah = !failedToGetToken()`, but I agree the `if` test is a bit clearer.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
```diff
@@ -24,6 +24,7 @@
     private static final long HE_TOKEN_TIMEOUT_MILLISECONDS = 10_000; // Timeout is a best guess
     // at what would be long enough to succeed and short enough for users to wait it out if
     // it should fail/time out, and do a normal flow
+    private static boolean heWasExecutedAndReceivedSuccessResponse = false;
```
vlarsen: When will this ever get back to `false`? When the app is closed and restarted? So it might survive for instance a SIM change?

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
SerhiiBovtriuk: The logic had changed. Please, check the new version out.

connect/src/com/telenor/connect/headerenrichment/HeLogic.java
SerhiiBovtriuk: The logic had changed. Please, check the new version out.

PR #203, opened 2020-04-17
Make instant verification optional

SerhiiBovtriuk:


SerhiiBovtriuk:
- Automatic instant verification can be switched on/off while initialising the sdk
- If switched on - instant verification will happen on login button initialisation
- There is an option to trigger instant verification manually

PR #204, opened 2020-04-17
Add extra information to connect api requests

SerhiiBovtriuk:


paulrene (commented):
I think what was outlined was a change of the User-Agent header in the request.

PR #210, opened 2020-11-11
Update HE endpoint

SerhiiBovtriuk:


oyvang (approved):

paulrene (approved):

connect/src/com/telenor/connect/utils/ConnectUrlHelper.java
```diff
@@ -24,7 +24,7 @@
 
     public static final String ACTION_ARGUMENT = "com.telenor.connect.ACTION_ARGUMENT";
     public static final String OAUTH_PATH = "oauth";
-    private static final String HE_TOKEN_API_BASE_PATH = "id/extapi/v1/header-enrichment-token/";
+    private static final String HE_TOKEN_API_BASE_PATH = "v2/extapi/v1/header-enrichment-token/";
```
paulrene, line 27: This new endpoint will not be available on the old connect hosts.

SerhiiBovtriuk: @paulrene it was confirmed that we are good to go with this change.

vlarsen: so, is it v2 or v1?

SerhiiBovtriuk: @vlarsen this is a trap for potentials hackers.

PR #212, opened 2021-01-11
Add device id functionality

SerhiiBovtriuk:


connect/src/com/telenor/connect/utils/ConnectUrlHelper.java
```diff
@@ -52,6 +52,9 @@ public static synchronized Uri getAuthorizeUri(
         if (heToken != null) {
             parameters.put("telenordigital_he_token", heToken);
         }
+        if (deviceId != null) {
+            parameters.put("did", deviceId);
```
paulrene, line 56: Call it "telenordigital_did" as we need to do that to show this is an extension to the oauth protocol.

paulrene (approved):

PR #224, opened 2021-03-12
Fix the issue when tokens are removed without need

SerhiiBovtriuk:


connect/src/com/telenor/connect/ConnectSdk.java
```diff
@@ -592,18 +595,45 @@ private static boolean fetchBooleanProperty(ApplicationInfo appInfo, String prop
...
-            wellKnownConfig = null;
-            // This operation may (and will) cause UI delay for the user. This is done on purpose.
-            // Cleaning of invalid idp configuration must be ensured.
-            lastSeenWellKnownConfigStore.clearSynchronously();
-            connectStore.clearSynchronously();
+            String wellKnownConfigIssuer = wellKnownConfig.getIssuer();
+            if (!expectedIssuer.equals(wellKnownConfigIssuer)) {
+                wellKnownConfig = null;
+                lastSeenWellKnownConfigStore.clearSynchronously();
+                connectStore.clearSynchronously();
```
ulfcarlin, line 635: We are testing if connectStore is not null higher up in this method. Could connectStore be null here?

SerhiiBovtriuk: It shouldn't, however it's safer to keep a check to prevent possible issues in future.

ulfcarlin (approved):

